PR #15, opened 2022-01-23
Feat: send the output of the runner to the server 

moreirathomas:
## Description

Collects the request's records as they are made concurrently and build a report.
Send the report to a server.

## Changes

Abstract a lot of logic and passing around of params inside `request.Requester`:
- package `request` becomes `requester`
- define the `Requester` struct configured with `config.Config`
- only expose `Requester.Run`, `Requester.Report` and `Requester.RunAndReport`


cmd/runner/main.go
```diff
@@ -44,9 +45,23 @@ func main() {
 	cfg := makeRunnerConfig()
 	fmt.Println(cfg)
 
-	rec := request.Do(cfg)
+	r := requester.New(requester.Options{
+		Concurrency: cfg.RunnerOptions.Concurrency,
+		Requests:    cfg.RunnerOptions.Requests,
+		Duration:    cfg.RunnerOptions.GlobalTimeout,
+		Timeout:     cfg.Request.Timeout,
+	})
```
moreirathomas, line 53: `config.Config` was not structured in a way that allowed the definition of `Requester` as a _doer_.
This interfacing is temporary. We need to discuss how to do this cleanly, suggestions more than welcomed.

cmd/runner/main.go
```diff
@@ -44,9 +45,23 @@ func main() {
...
 
-	fmt.Println("total:", len(rec))
+	ctx, cancel := context.WithTimeout(context.Background(), cfg.RunnerOptions.GlobalTimeout)
+	defer cancel()
+
+	r.Run(ctx, requester.Target{
+		Method: cfg.Request.Method,
+		URL:    cfg.Request.URL,
+	})
+	report := r.Collect()
```
moreirathomas, line 62: All of that can be hidden behind something like `Requester.RunAndReport(url)`.
The need to extract `ctx` comes from the pipeline nature of `Run` and `Collect`. It may be abstracted inside a single "execute" method.

To discuss.

GregoryAlbouy (approved):

c1-ra (approved):
Looks good to me :)

PR #16, opened 2022-01-23
dev: write unit tests for concurrency pattern

### LINKED ISSUE #14: Unexpectedly high amount of requests before global timeout
## Description

See https://github.com/benchttp/runner/pull/8#discussion_r790142374

GregoryAlbouy:
## Description

This PR adds unit tests for what used to be `request.Do` and `request.DoUntil`.

However, this couldn't have been done without a consequent refactoring first: see changes below.

It also fixes #14 in the process.

## Changes

- Make the semaphore implementation testable by uncoupling it from the request business (the http call based on url is a side effect that is hardly testable in a unit test)
- Make the semaphore implementation more generic so it handles both `Do` and `DoUntil` in a single `Do`, which makes one function less to test
- Simplify the semaphore implementation:
  - Use dedicated data structure (`record.SafeSlice`) to handle safe concurrent use: removes one chan, one wrapping goroutine, and the need to collect
  - Use fully tested package `golang.org/x/sync/semaphore`: less things to test, and removes the custom context handling consisting in a `select` wrapped in a `for` loop wrapped in a `func` (only the loop remains)

## Notes

> the http call based on url is a side effect that is hardly testable in a unit test

For this reason, I don't think we can write a proper unit test for `request.Do`, which internally makes an http request based on its given url.
I suggest we set up an end-to-end testing environment with Docker that uses our benchttp/cobaye to make such tests.


semimpl/do.go
```diff
@@ -0,0 +1,73 @@
...
+	"errors"
+	"log"
+	"sync"
+
+	"golang.org/x/sync/semaphore"
+)
+
+// Do concurrently executes callback at most maxIter times or until ctx is done
+// or canceled. Concurrency is handled leveraging the semaphore pattern, which
+// ensures at most numWorkers goroutines are spawned at se same time.
```
c1-ra, line 14: [nitpick] "... at the same time"

semimpl/do.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	return maxIter
+}
+
+func sanitizeCallback(callback func()) func() {
+	if callback == nil {
+		return func() {}
+	}
+	return callback
+}
```
c1-ra, line 63: [suggestion] 
Instead of correcting the values provided, maybe we could raise an error and inform the user?
It is not 100% necessary though since we make a summary of the values which have been actually used in the output, the user will know if it is not what they intended to input.

c1-ra: Or maybe just add a message in the output next to the value changed? ("Concurrent requests number  must be >= 1" for instance)

GregoryAlbouy: If we consider the fact the results are going to be uploaded, it's indeed a good idea to stop the process for any weird value.

Which values should be weird?
- `callback == nil`, because it would iterate doing literally nothing
- `numWorkers == 0`? 0 is not logical because we can't have 0 worker, but here we allow it to mean "no concurrency". We also allow `-1` because it's quite common, and at this point why not any negative number? (someone using -732 is likely not taking it seriously)
- `maxIter == 0`? Same

`-1` is often used for this purpose, but `0` is kind of convenient here because it's Go's default value for ints so it seems quite logical that when `-c` is not set, its default value matches "no concurrency" (same with `-r` / `maxIter`)

GregoryAlbouy: Alternatively, we can use a strict logic for `semimpl` (`numWorkers > 0` and `maxIter == -1 || maxIter > 0`) and handle the defaulting at a high level, in `cmd` or `request`.

semimpl/do_test.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/benchttp/runner/semimpl"
+)
+
+func TestDo(t *testing.T) {
+	t.Run("stop until maxIter is reached", func(t *testing.T) {
```
c1-ra, line 12: [nitpick]
I think "stop when maxIter is reached" would be a little clearer.

semimpl/do_test.go
```diff
@@ -0,0 +1,177 @@
...
+			)
+		}
+	})
+
+	t.Run("stop on context cancel", func(t *testing.T) {
+		var (
+			timeout    = 50 * time.Millisecond
+			interval   = 10 * time.Millisecond
+			numWorkers = 1
+			maxIter    = 0 // infinite iterations
```
c1-ra, line 62: [suggestion]
Is there a reason you prefer to use `0` for maxIter instead of `nil`?
It makes a little more sense to me that `nil` would mean that there is no maximum, when 0 seems strange (but it may be because I'm not used to Go).

semimpl/do_test.go
```diff
@@ -0,0 +1,177 @@
...
+	})
+
+	t.Run("stop on context timeout", func(t *testing.T) {
+		var (
+			timeout    = 50 * time.Millisecond
+			interval   = timeout / 5
+			numWorkers = 1
+			maxIter    = 0 // infinite iterations
+
+			margin      = 15 * time.Millisecond
```
c1-ra, line 36: [question]
How did you decide how much should the margin be?
This fails on my computer, I need at least to replace this by 20 for this test to pass.

semimpl/do_test.go
```diff
@@ -0,0 +1,177 @@
...
+	})
+
+	t.Run("stop on context cancel", func(t *testing.T) {
+		var (
+			timeout    = 50 * time.Millisecond
+			interval   = 10 * time.Millisecond
+			numWorkers = 1
+			maxIter    = 0 // infinite iterations
+
+			margin      = 15 * time.Millisecond
```
c1-ra, line 64: This one does not pass either on my computer except if I use a margin of 20.

semimpl/do_test.go
GregoryAlbouy: Yup, `int` can't be `nil`. Only a few non-primitives types can: maps, slices (not arrays), funcs, interfaces, pointers (I'm probably missing some). They have instead a default value: `0` for numbers, `""` for strings, `false` for bools...
Not having to deal with `nil` is actually quite handy, but the downside is you can't know whether a value has been initialized or defaulted.

semimpl/do_test.go
GregoryAlbouy: 100% empirically, I'll update it.

semimpl/do_test.go
```diff
@@ -0,0 +1,177 @@
...
+			curr, next := minof(group), maxof(groups[groupIndex+1])
+			if interval := next - curr; interval < minIntervalBetweenGroups {
+				t.Errorf(
+					"unexpected interval between groups: exp > %dms, got %dms",
+					minIntervalBetweenGroups.Milliseconds(), interval.Milliseconds(),
+				)
+				t.Log(durations)
+			}
+		}
+	})
```
c1-ra, line 135: [nitpick]
If I understand correctly, `durations` is the list of time at which each request has been done?
It is nitpicking, but isn't it a specific time rather than a duration (I get that this time = the duration from the start, but "durations" makes it sounds like it is the duration each request took, which is not the case here I think)? Could we rename it?

semimpl/do_test.go
c1-ra: Ah, I forgot this... Ok, 0 it is then 😄

semimpl/do_test.go
```diff
@@ -0,0 +1,181 @@
...
+			maxIter    = 12
+
+			durations                = make([]time.Duration, 0, maxIter)
+			maxIntervalWithinGroup   = 10 * time.Millisecond
+			minIntervalBetweenGroups = 30 * time.Millisecond
+		)
+
+		start := time.Now()
+		semimpl.Do(context.Background(), numWorkers, maxIter, func() {
+			durations = append(durations, time.Since(start))
```
GregoryAlbouy, line 100: Note: this assignation is technically not thread-safe. However with only concurrent 3 workers the odds of a race condition is probably negligible
If that bothers someone I'll update it.

semimpl/do_test.go
GregoryAlbouy: I also went for `times` at first, but I was bothered having a `[]time.Duration` named `times`, which I expect to be `[]time.Time`.
We can still work on a more specific name, or at worst explicit it in a comment 🤷

semimpl/do_test.go
```diff
@@ -0,0 +1,177 @@
...
+					maxIntervalWithinGroup.Milliseconds(), interval.Milliseconds(),
+				)
+				t.Log(durations)
+			}
+
+			// check durations between distinct groups are spaced
+			if groupIndex == len(groups)-1 {
+				break
+			}
+			curr, next := minof(group), maxof(groups[groupIndex+1])
```
GregoryAlbouy, line 126:
```suggestion
			curr, next := maxof(group), minof(groups[groupIndex+1])
```

semimpl/do_test.go
c1-ra: I don't mind having a `[]time.Duration` named `times` but I understand if you do 😁 
A comment also works if we do not come up with a more specific name.

semimpl/do_test.go
GregoryAlbouy, line 126:
```suggestion
			curr, next := maxof(group), minof(groups[groupIndex+1])
```

c1-ra: Nice catch!

semimpl/do.go
c1-ra: I would go for the 2nd option (`numWorkers > 0` `and maxIter > 0 || maxIter == -1`) and handle defaulting at a higher level if others agree.

I don't know for a safeguard, because I have no idea what would be a reasonable number and I think it would be the user responsibility to know roughly how much they can ask from their machine.
Or an option would be to add an arbitrary limit, but I feel like it should still be possible to explicitly override it.

semimpl/do_test.go
c1-ra: It is very unlikely there is a problem, but I feel like it would be cleaner to make it thread-safe.
I guess it could also be considered to be overkill though 🤷‍♀️

c1-ra:
Except for my small comments, nitpicks and questions, it looks good to me.
Plus, I think the refactoring indeed makes the code simpler (even though it makes my first PR obsolete 😝).

semimpl/do.go
GregoryAlbouy: I realize there's more to do in terms of error handling: in addition to `semimpl.Do`, we must also handle invalid url or failed requests in `requesst.Do`. For that reason I suggest we create a dedicated issue for global error handling in the app.
It's not a huge piece of work, but it'll keep this PR well scoped and without behavioural changes.

moreirathomas: > I realize there's more to do in terms of error handling: in addition to semimpl.Do, we must also handle invalid url or failed requests in requesst.Do. For that reason I suggest we create a dedicated issue for global error handling in the app.

I don't think it should. The function it is passed must handle errors itself. `semimpl.Do` ùust not know what behavior it is wrapping. Something like: 

```go
semimpl.Do(ctx, concurrency, requests, func() {
	rec, err := doRequest(url, timeout)
	if err != nil {
		fails += 1
		return
	} else {
		records.Append(rec)
	}
})
```

Anyway, it's outside the scope of this PR :)

GregoryAlbouy: Last comment: obviously, that's exactly what I meant by _handle invalid url or failed requests **IN** `request.Do`_ and by _global error handling in the app_.

> `callback == nil` should be a compile time error (although `func()` accepts `nil`)

Should we rewrite the Go compiler then? 😛 

So in the end it seems we all agree on the error handling specs, time to write the issue 🚀

semimpl/do.go
```diff
@@ -0,0 +1,73 @@
...
+
+func handleContextError(err error) {
```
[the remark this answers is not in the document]
GregoryAlbouy: We could just do nothing: it's not _really_ an error as it won't happen unexpectedly, it's just how a `context` tells us it has timeout / been canceled, and we have full control over it on its creation.
The real error handling here is `default: log.Fatal(err)` (to be improved later), the rest are non-errors.

semimpl/do.go
```diff
@@ -0,0 +1,73 @@
...
+	switch {
+	case err == nil:
```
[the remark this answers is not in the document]
GregoryAlbouy: I know, but regarding the function itself I was a bit bothered omitting a case by making an assumption on how it is used 😕

semimpl/do.go
```diff
@@ -0,0 +1,73 @@
...
+// ensures at most numWorkers goroutines are spawned at the same time.
+func Do(ctx context.Context, numWorkers, maxIter int, callback func()) {
+	numWorkers = sanitizeNumWorkers(numWorkers)
+	maxIter = sanitizeMaxIter(maxIter)
+	callback = sanitizeCallback(callback)
+
+	sem := semaphore.NewWeighted(int64(numWorkers))
+	wg := sync.WaitGroup{}
+
+	for i := 0; i < maxIter || maxIter == 0; i++ {
```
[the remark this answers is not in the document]
GregoryAlbouy: I don't think it's possible with making things overly complex regarding what we're trying to achieve 🤔 

My 2 cents:
- If the impact on performance is negligible, then 🤷‍♂️ 
- If it's notable, then it's a good thing we have the same implementation for both strategies.

semimpl/do.go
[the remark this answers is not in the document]
GregoryAlbouy: Oh, maybe you meant what do we do instead of `log.Fatal`?
Well I just checked, this case should never happen: the only error values ever returned by `sem.Acquire` are `nil` and the `context` errors which are handled here.

semimpl/do.go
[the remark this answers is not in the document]
moreirathomas: Fair enough 👌

moreirathomas (dismissed):

semimpl/do.go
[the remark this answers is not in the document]
c1-ra: I would say we can remove this case too.

c1-ra (dismissed):
I'm ok to merge it too!

moreirathomas (approved):

moreirathomas (approved):

c1-ra (approved):
Looks good :)

PR #21, opened 2022-01-26
feat: parse input config file

GregoryAlbouy:
## Description

This PR adds config file parsing capabilites to the runner.
It does not follow all the specs from the issue: it will be properly updated and the current one will be rescoped to a MVP milestone.
Until then, you can see the config file examples in `test/testdata/config` for the specs.

### Examples

```
# Build
make build

---

# CLI changes
./bin/benchttp -concurrency 10 -requests 100 -timeout 1s -globalTimeout 30s -url "http://localhost:9999"

# With default config path (`./.benchttp.yml`)
./bin/benchttp

# With custom config path
./bin/benchttp -configFile ./test/testdata/config/benchttp.yml

# Same with CLI override
./bin/benchttp -configFile ./test/testdata/config/benchttp.yml -requests 1000 -url "http://example.com"
```

## Changes

- New cli option `-configFile` (default `"./.benchttp.yml"`) to read the value as a config file
- New cli option `-url` (default `""`) as it became optional (can be set from file)
- Change cli option names to match config file
- Config from the CLI override the one from config files

## Todo

- [x] Rework the global defaulting process
- [x] Add cfg file defaulting sequence `.yml` > `.yaml` > `.json`
- [x] ~Error handling~
- [x] Refactor `config.Config`
- [x] Update docs


request/do.go
```diff
@@ -32,11 +33,22 @@ func doRequest(url string, timeout time.Duration) record.Record {
...
-
-	semimpl.Do(ctx, concurrency, requests, func() {
-		records.Append(doRequest(url, timeout))
+func Do(cfg config.Config) []record.Record {
+	var (
+		uri        = cfg.Request.URL.String()
+		numWorker  = cfg.RunnerOptions.Concurrency
+		numRequest = cfg.RunnerOptions.Requests
+		reqTimeout = cfg.Request.Timeout
+		gloTimeout = cfg.RunnerOptions.GlobalTimeout
```
moreirathomas, line 42: Ew that's not sexy.

I really think `duration` was better, one of the reason is exactly this one, struggling to differentiate things with the same name by adding more characters to it in unsatisfying ways.

`duration` really means "the entire duration the benchmark should run for, leave blank to default to whatever we choose was the acceptable max duration other than infinity".

request/do.go
```diff
@@ -32,11 +33,22 @@ func doRequest(url string, timeout time.Duration) record.Record {
...
+	)
+
+	records := record.NewSafeSlice(numRequest)
+
+	ctx, cancel := context.WithTimeout(context.Background(), gloTimeout)
+	defer cancel()
```
moreirathomas, line 48: If we were going the go way by passing `context.Context` from the top level process aka main to here, this is a step backward.
Because now the context is not a red thread passed from modules to modules but something one module decide it is.
What was the rationale behind this change ?

cmd/runner/main.go
```diff
@@ -1,56 +1,76 @@
...
+		// config file is not mandatory, other errors are critical
+		log.Fatal(err)
+	}
+
+	cliConfig := config.New(url, requests, concurrency, timeout, globalTimeout)
+
+	cfg := config.Merge(fileConfig, cliConfig)
+
+	return cfg
+}
```
moreirathomas, line 65: This is not a very pleasing api. It works but we can do better ! Suggestion: 

How I want to use it as a main ?

```go
func main() {
	cfg, err := config.Parse(file)
	if err != nil && !errors.Is(err, os.ErrNotExist) {
		log.Fatalf("could not configure the runner: %s", err)
	}

	cfg = config.Merge(cfg, args)
	// ...
}
```

We can even go further, as main, I don't really care about the parsing and merging steps. So we can have an api like :
```go
// package config

func Parse(file string) (Config, error)

func Merge(cfg Config, args Type?) (Config, error)

func ParseAndMerge(file string, args Type?) (Config, error) {
	return Merge(Parse(file), args)
}
```
Only `ParseAndMerge` could be public if it is the usecase in the end.

This obviously means packages `config` and `configparser` should be merge. I'll have another comment about that.

cmd/runner/main.go
```diff
@@ -1,56 +1,76 @@
...
-	flag.IntVar(&requests, "r", DefaultRequests, "Number of requests to run, use duration as exit condition if omitted")
-	flag.DurationVar(&duration, "d", DefaultDuration, "Duration of test, in seconds")
-	flag.DurationVar(&timeout, "t", DefaultTimeout, "Timeout for each http request, in seconds")
+func parseArgs() {
+	flag.StringVar(&configFile, "configFile", findFile(defaultConfigFiles), "Config file path")
+	flag.StringVar(&url, "url", "", "Target URL to request")
+	flag.IntVar(&concurrency, "concurrency", 0, "Number of connections to run concurrently")
+	flag.IntVar(&requests, "requests", 0, "Number of requests to run, use duration as exit condition if omitted")
+	flag.DurationVar(&timeout, "timeout", 0, "Timeout for each http request")
+	flag.DurationVar(&globalTimeout, "globalTimeout", 0, "Duration of test")
```
moreirathomas, line 37: Why do the flags needed to lose their shorthand ? it is said it was to match config file but to me it seems like the code could still really work with shorthand flags and be more dev-friendly

Also I still think expecting `url` as an argument and not a flag is better. The real reason is : it is the only on required as everything else can be defaulted in a sane and predictable manner (by the end user I mean, "yeah if I don't provide a number of workers it'll just be the default", not true for url).
Plus when we'll have others commands (ie `benchttp login`), it will make more sense for the command `run`/`test` to be passed an argument (`benchttp test http://myapp.foo/endpoint`).

moreirathomas (changes requested):
Overall nice, but I think we need to build upon that to have a nicer api for this module.

Note : packages `config` and `configparser` **need** to be merged. When you look at the imports what really is going on is `configparser` is an internal of `config` dealing with parsing yaml and json files. Also their internals leak into `main` and end up inside `makeRunnerConfig`.

What I suggest is : 

```txt
├── cmd
│   └── runner
│       └── main.go
├── config
│   ├── config.go
│   ├── merge.go
│   ├── parse.go
│   └── parser
│       ├── yaml.go
│       └── json.go

```
Files are to convey the purpose of each package, in reality json and yaml can be generic (see other comment below) and the sub package could really be a simple file inside `config` (if we go the route of `config.ParseAndMerge` or similar).

cmd/runner/main.go
GregoryAlbouy: CLI options reflect the config file options, they need to share the same name for consistency. Consistency is more dev-friendly than shortcuts. See [TypeScript](https://www.typescriptlang.org/docs/handbook/compiler-options.html) or [Docker](https://docs.docker.com/engine/reference/commandline/cli/): no shortcuts needed.
We can still add shorthands later though, but this is the least of our priorities.

`url` can't be required in the CLI because it can be set in the config file, in which case we don't want to override it. This is how it was specced: we can still re-discuss it, but I suggest not here. Error handling for an empty url won't be an issue.

When we have other commands, we'll update the API as needed – now is not the time.

request/do.go
GregoryAlbouy: Nonsense, why would one require to pass the context from main? What I said was using context is a handy pattern specifically designed to achieve what we were doing manually.
Reason is obvious: start the global timout timer closer to when the requests actually starts.

cmd/runner/main.go
GregoryAlbouy: Strongly disagree: `Config` has nothing to do with reading and parsing files and should know nothing about that business. When `request` imports `config`, it should not import a parsing module.

request/do.go
GregoryAlbouy: This corresponds to the specs we agreed one. We can still re-discuss them, but I suggest not here.

moreirathomas: I'm fine with that for now but i think we can still re-discuss the specs here as they are fuzzy at best. See:

> It does not follow all the specs from the issue: it will be properly updated and the current one will be rescoped to a MVP milestone.

configparser/parser.go
```diff
@@ -0,0 +1,43 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type extension string
+
+const (
+	extYML  extension = ".yml"
+	extYAML extension = ".yaml"
+	extJSON extension = ".json"
+)
```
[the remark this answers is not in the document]
GregoryAlbouy: It's an enum. Enums are a well-suited pattern for a set of limited known values. They provide a better visibility on what values are accepted (rather than having to look inside the code in a switch statement) and self-documenting typing for function parameters: an extension is not any string.
The fact the compiler does not enforce that typing is not a sufficient reason to leave the other benefits.

`default` case *does* happen in case you provide a file with an invalid extension.

request/do.go
GregoryAlbouy: What I meant here is that I didn't implement _all_ the features requested. Not that I would do just as it pleases me.
Not discussing that here.

cmd/runner/main.go
moreirathomas: I see your point, but if we look at what the purpose of each package we have this:

I use `configparser.Parse` to get a `config.Config`. I use `config.Parse` to get a `config.Config` and `config.Merge` to get another `config.Config`. 
These are the only functionalities offered by the 2 packages.

`request` import `config` to simply type the input of `request.Do`, it does not require any functionality.

`config` really just bundles params into a nice structure for use by other parts of the system. So I don't really think there is distinction between the reading and parsing config file and structuration of the config object.

configparser/parser.go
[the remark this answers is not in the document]
moreirathomas: Can't you assert the type is right when you convert? If no then i was wrong on the `default` case. It thought it was what was going on.

cmd/runner/main.go
GregoryAlbouy: `Config` defines the behaviour of the whole runner. It's a domain entity that could possibly be used anywhere in the app. It's in `configparser`, it's in `request`, it'll also be in `output` (whatever the name) because we want to post that config as well.
`configparser` provides low-level utility to get a `Config` out of a file. `Config` must not be coupled with that.

I think we both made our points, however if anyone else has an opinion on that topic, they're welcome.

configparser/parser.go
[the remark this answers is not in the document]
GregoryAlbouy: The typing doesn't even matter here: at runtime the switch tries to match a value. If that value is `.bruvbzo`, then it's none of the cases, then its `default.`
Type safety can't do much at runtime, at this point it's error handling.

cmd/runner/main.go
moreirathomas: I do not agree, it is not an _entity_ if it does not have behaviours attached. There is no `Config.Do` procedure anywhere. It's just a bag of parameters for other entities/actors (the requester, the thing that will send the results, ...).

All that is exported form `config` is its struct and `New` (or `NewButMerge`).
What it really is compared to an entity, is a simple value object. Same as `record` in fact.

configparser/parser.go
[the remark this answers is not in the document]
moreirathomas: Ok then mb

cmd/runner/main.go
GregoryAlbouy: Side note: no thumbs down. It's an unconstructive way to express disagreement and should be avoided.

c1-ra: When in doubt, I would tend to separate the functionalities in different packages.
I feel like even if we realize later that it is not ideal, it will be easier to regroup different functionalities in the same package than the opposite.

configparser/parse_test.go
```diff
@@ -0,0 +1,109 @@
...
+			gotURL, expURL := gotCfg.Request.URL, expCfg.Request.URL
+
+			// compare *url.URLs separately, as they contain unpredictable values
+			// they need special checks
+			if !sameURL(gotCfg.Request.URL, expCfg.Request.URL) {
+				t.Errorf(
+					"unexpected parsed URL: exp %v, got %v",
+					expCfg.Request.URL, gotCfg.Request.URL,
+				)
+			}
```
c1-ra, line 47: Wouldn't it be simpler to use `gotURL `and `expURL `here instead of `gotCfg.Request.URL` and `expCfg.Request.URL`?

cmd/runner/main.go
```diff
@@ -1,56 +1,76 @@
...
+}
 
-	flag.IntVar(&concurrency, "c", DefaultConcurrency, "Number of connections to run concurrently")
-	flag.IntVar(&requests, "r", DefaultRequests, "Number of requests to run, use duration as exit condition if omitted")
-	flag.DurationVar(&duration, "d", DefaultDuration, "Duration of test, in seconds")
-	flag.DurationVar(&timeout, "t", DefaultTimeout, "Timeout for each http request, in seconds")
+func parseArgs() {
+	flag.StringVar(&configFile, "configFile", findFile(defaultConfigFiles), "Config file path")
+	flag.StringVar(&url, "url", "", "Target URL to request")
+	flag.IntVar(&concurrency, "concurrency", 0, "Number of connections to run concurrently")
```
c1-ra, line 34: Shouldn't we rather use 1 for default concurrency? Since no concurrency = 1 worker at a time.

GregoryAlbouy: 5d4fd1d59e78ef3d6a5bedef26e84be84bcccfcf
> Rework config defaulting
>
> - use zero values for cli defaults (so they don't override when not set)
> - use config.Default() for default config

We need to distinguish the CLI default values from the config defaults. The reason is we override the options from config file with the CLI options _when they are set_. If we use `1` as CLI default, it will be considered set to `1` by default and the config file value will always be ignored.

c1-ra (approved):

config/config.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Default returns a default config that is safe to use.
+func Default() Config {
+	return defaultConfig
+}
+
+// Merge returns a Config after a base Config overridden by all non-zero values
+// of override.
+func Merge(base, override Config) Config {
```
GregoryAlbouy, line 58: This need unit tests

moreirathomas (approved):

c1-ra (approved):
Looks good to me! 😃

PR #23, opened 2022-01-30
feat: implement request tracer

GregoryAlbouy:
## Description

Implement and add profiling features to the request benchmarks.

Instead of measuring the total duration only, we now have access to a timestamp for each step of an http request.
This will allow for more insightful data for the user.

### Example

- Local request (`http`)
  ```
  ./bin/benchttp -requests 1 -url "http://localhost:9999"
  ```
  
  <details>
  <summary>Output</summary>
  <pre>
  [
    {
      "Cost": 0,
      "Code": 200,
      "Bytes": 0,
      "Error": null,
      "Profile": [
        {
          "Name": "GetConn",
          "Time": 70
        },
        {
          "Name": "DNSStart",
          "Time": 38155
        },
        {
          "Name": "DNSDone",
          "Time": 836820
        },
        {
          "Name": "ConnectStart",
          "Time": 866851
        },
        {
          "Name": "ConnectDone",
          "Time": 1072322
        },
        {
          "Name": "ConnectStart",
          "Time": 1125989
        },
        {
          "Name": "ConnectDone",
          "Time": 1257758
        },
        {
          "Name": "GotConn",
          "Time": 1290506
        },
        {
          "Name": "WroteHeaders",
          "Time": 1358900
        },
        {
          "Name": "WroteRequest",
          "Time": 1361062
        },
        {
          "Name": "GotFirstResponseByte",
          "Time": 1473155
        },
        {
          "Name": "PutIdleConn",
          "Time": 1528966
        }
      ]
    }
  ]
  </pre>
  </details>

- Remote request (`https`)
  ```
  ./bin/benchttp -requests 1 -url "https://example.com"
  ```
  
  <details>
  <summary>Output</summary>
  <pre>
  [
    {
      "Cost": 0,
      "Code": 200,
      "Bytes": 1256,
      "Error": null,
      "Profile": [
        {
          "Name": "GetConn",
          "Time": 74
        },
        {
          "Name": "DNSStart",
          "Time": 42508
        },
        {
          "Name": "DNSDone",
          "Time": 10587412
        },
        {
          "Name": "ConnectStart",
          "Time": 10655260
        },
        {
          "Name": "ConnectDone",
          "Time": 89404642
        },
        {
          "Name": "TLSHandshakeStart",
          "Time": 89554213
        },
        {
          "Name": "TLSHandshakeDone",
          "Time": 414801212
        },
        {
          "Name": "GotConn",
          "Time": 415041317
        },
        {
          "Name": "WroteHeaders",
          "Time": 415092642
        },
        {
          "Name": "WroteRequest",
          "Time": 415092837
        },
        {
          "Name": "GotFirstResponseByte",
          "Time": 497572965
        }
      ]
    }
  ]
  </pre>

### DIFF
--- a/requester/requester.go
+++ b/requester/requester.go
@@ -64,7 +64,7 @@ type Record struct {
 	Time   time.Duration `json:"time"`
 	Code   int           `json:"code"`
 	Bytes  int           `json:"bytes"`
-	Error  error         `json:"error"`
+	Error  error         `json:"error,omitempty"`
 	Events []Event       `json:"events"`
 }
 
--- a/requester/requester.go
+++ b/requester/requester.go
@@ -16,17 +16,24 @@ type Requester struct {
 
 	config config.Config
 	client http.Client
+	tracer *tracer
 }
 
 // New returns a Requester configured with specified Options.
 func New(cfg config.Config) *Requester {
+	tracer := newTracer()
 	return &Requester{
 		records: make(chan Record, cfg.RunnerOptions.Requests),
 		config:  cfg,
+		tracer:  tracer,
 		client: http.Client{
-			// Timeout includes connection time, any redirects, and reading the response body.
+			// Timeout includes connection time, any redirects, and reading
+			// the response body.
 			// We may want exclude reading the response body in our benchmark tool.
 			Timeout: cfg.Request.Timeout,
+
+			// tracer keeps track of all events of the current request.
+			Transport: tracer,
 		},
 	}
 }
@@ -54,10 +61,11 @@ func (r *Requester) Run() Report {
 // response body, invalidating the entire response, as it is not a remote
 // server error.
 type Record struct {
-	Time  time.Duration `json:"time"`
-	Code  int           `json:"code"`
-	Bytes int           `json:"bytes"`
-	Error error         `json:"error"`
+	Time   time.Duration `json:"time"`
+	Code   int           `json:"code"`
+	Bytes  int           `json:"bytes"`
+	Error  error         `json:"error"`
+	Events []Event       `json:"events"`
 }
 
 func (r *Requester) record() {
@@ -83,8 +91,9 @@ func (r *Requester) record() {
 	}
 
 	r.records <- Record{
-		Code:  resp.StatusCode,
-		Time:  time.Since(sent),
-		Bytes: len(body),
+		Code:   resp.StatusCode,
+		Time:   time.Since(sent),
+		Bytes:  len(body),
+		Events: r.tracer.events,
 	}
 }
--- a/requester/tracer.go
+++ b/requester/tracer.go
@@ -7,14 +7,14 @@ import (
 	"time"
 )
 
-// Event is a stage of an outgoing http request associated with a timestamp.
+// Event is a stage of an outgoing HTTP request associated with a timestamp.
 type Event struct {
 	Name string        `json:"name"`
 	Time time.Duration `json:"time"`
 }
 
 // tracer is a http.RoundTripper to be used as a http.Transport
-// that records the events of an outgoing http request.
+// that records the events of an outgoing HTTP request.
 type tracer struct {
 	start  time.Time
 	events []Event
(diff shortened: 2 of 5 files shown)


requester/tracer.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+		WroteRequest: func(httptrace.WroteRequestInfo) {
+			p.addEvent("WroteRequest")
+		},
+		GotFirstResponseByte: func() {
+			p.addEvent("GotFirstResponseByte")
+		},
+		PutIdleConn: func(error) {
+			p.addEvent("PutIdleConn")
+		},
```
c1-ra, line 72: [question] It took me some time to understand why we have parameters we don't use.
It is just because it is in the signature of the hook methods, right?
And it is provided **just in case** we wanted to use it in the method called at this time?

c1-ra (approved):
Nice :+1:

requester/tracer.go
GregoryAlbouy: Yes exactly! The signatures are specific to the hooks and we have to comply with them (even if we don't actually use them).

moreirathomas: Question: is it better to completely avoid the param name or explicitly discard it with `_`:
- `func(string) {...}` vs
- `func(_ string) {...}`
I know that `_` is the way to discard so its obvious but I'd like your thoughts on that.

GregoryAlbouy: `_` is necessary if one wants a ignore some parameter(s) amongst others:
```go
// unexpected
func Do(name string, int, isCool bool) {}
                  // ^^^ this is a bool param named "int"

// ok
func Do(name string, _ int, isCool bool) {}
```

Being unnecessary in our case, and parameters being there only to comply with the required signature, I prefer the more concise version to this:
```go
ConnectDone: func(_, _ string, _ error) {
    p.addEvent("ConnectDone")
},
```

requester/requester.go
```diff
@@ -16,17 +16,24 @@ type Requester struct {
...
+	tracer *tracer
 }
 
 // New returns a Requester configured with specified Options.
 func New(cfg config.Config) *Requester {
+	tracer := newTracer()
 	return &Requester{
 		records: make(chan Record, cfg.RunnerOptions.Requests),
 		config:  cfg,
+		tracer:  tracer,
```
moreirathomas, line 28: Suggestion: inline `newTracer()` here ?

requester/requester.go
```diff
@@ -54,10 +61,11 @@ func (r *Requester) Run() Report {
...
 // server error.
 type Record struct {
-	Time  time.Duration `json:"time"`
-	Code  int           `json:"code"`
-	Bytes int           `json:"bytes"`
-	Error error         `json:"error"`
+	Time   time.Duration `json:"time"`
+	Code   int           `json:"code"`
+	Bytes  int           `json:"bytes"`
+	Error  error         `json:"error"`
```
moreirathomas, line 67: Not really in the scope of this PR but could be added because we could avoid to open a new one to do so:
Error could be omitted if empty in json.
```suggestion
	Error  error         `json:"error,omitempty"`
```

requester/tracer.go
```diff
@@ -0,0 +1,85 @@
+package requester
+
+import (
+	"crypto/tls"
+	"net/http"
+	"net/http/httptrace"
+	"time"
+)
+
+// Event is a stage of an outgoing http request associated with a timestamp.
```
moreirathomas, line 10: nitpick: uppercase "HTTP"?

requester/tracer.go
```diff
@@ -0,0 +1,85 @@
...
+// RoundTrip implements http.RoundTripper. It attaches the client trace
+// to the request context and calls http.DefaultTransport.RoundTrip
+// with the new created request.
+func (p *tracer) RoundTrip(r *http.Request) (*http.Response, error) {
+	ctx := httptrace.WithClientTrace(r.Context(), p.trace())
+	return http.DefaultTransport.RoundTrip(r.WithContext(ctx))
+}
+
+// trace returns a http.ClientTrace that timestamps and records the events
+// of an outgoint http requests.
```
moreirathomas, line 32:
```suggestion
// of an outgoing HTTP request.
```

moreirathomas (dismissed):
Small misc nitpicks.
Otherwise good !

requester/requester.go
GregoryAlbouy: I think it wouldn't work: we need the same `tracer` as the one used in `Requester.client.Transport` (L36).

We could do without any `Requester.tracer` however, retrieving the actual tracer it via `Requester.client.Transport.(*tracer)` assertion, but I feel the current approach is cleaner and safer 🤔

moreirathomas (approved):

c1-ra (approved):
:+1:

PR #25, opened 2022-01-30
fix: do not ignore concurrency from config file

### LINKED ISSUE #24: config.Merge overwrites file options with defaults if cli args have not been provided
## Description

I run the binary without providing any arguments and uses a custom config file `./.benchttp.yml`.
My custom config:

```yml
request:
  method: GET
  url: http://localhost:9999?delay=10ms
  queryParams:
    fib: 5
  timeout: 2s

runnerOptions:
  requests: 9999
  concurrency: 99
  globalTimeout: 2s
```

As I provided all options, I expect `config.Merge` to overwrite the default with everything written here. However by adding some logs inside `makeRunnerConfig`, this is what I get:

```json
file {
  "Request": {
    "Method": "GET",
    "URL": {
      "Scheme": "http",
      "Opaque": "",
      "User": null,
      "Host": "localhost:9999",
      "Path": "",
      "RawPath": "",
      "ForceQuery": false,
      "RawQuery": "delay=10ms\u0026fib=5",
      "Fragment": "",
      "RawFragment": ""
    },
    "Timeout": 2000000000
  },
  "RunnerOptions": {
    "Requests": 9999,
    "Concurrency": 99,
    "GlobalTimeout": 2000000000
  }
}
cli {
  "Request": {
    "Method": "GET",
    "URL": {
      "Scheme": "",
      "Opaque": "",
      "User": null,
      "Host": "",
      "Path": "",
      "RawPath": "",
      "ForceQuery": false,
      "RawQuery": "",
      "Fragment": "",
      "RawFragment": ""
    },
    "Timeout": 10000000000 <- defaulted by config.Default(), expected zero value
  },
  "RunnerOptions": {
    "Requests": 0,
    "Concurrency": 1, <- defaulted by config.Default(), expected zero value
    "GlobalTimeout": 30000000000 <- defaulted by config.Default(), expected zero value
  }
}
merged {
  "Request": {
    "Method": "GET",
    "URL": {
      "Scheme": "http",
      "Opaque": "",
      "User": null,
      "Host": "localhost:9999",
      "Path": "",
      "RawPath": "",
      "ForceQuery": false,
      "RawQuery": "delay=10ms\u0026fib=5",
      "Fragment": "",
      "RawFragment": ""
    },
    "Timeout": 10000000000 <- wrongly overwritten, expected config file to not be overwritten
  },
  "RunnerOptions": {
    "Requests": 9999,
    "Concurrency": 1, <- wrongly overwritten, expected config file to not be overwritten
    "GlobalTimeout": 30000000000 <- wrongly overwritten, expected config file to not be overwritten
  }
}
```

## Suggestions

I suspect the problem lies in the steps the merge follows:
- read flags, default to zero-values if not given
- call `config.New` with flag-parsed values, internally `MergeDefault` overwrites the zero-values, the cli _has provided_ values
- call `config.Parse` to get the config file values, works as expected
- call `config.Merge` which overwrite file's values by cli's ones, but the cli one has been "corrupted" already

I have no immediate solution to suggest as I did not think enough about that. But I suspect an incremental build of the config would be easier to tackle when compare to handling a merge where so many null values can come from many different places (args, file and default).

Thinking is required ! 
For now I'll just pass args from the cli to short circuit the defaulting behavior.

GregoryAlbouy:
## Description

Fixes #24 

## Changes

- `config.New` does not return a merged config with `config.Default` anymore
- Updated unit tests accordingly
- Updated docs accordingly

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,8 +34,14 @@ func (cfg Config) String() string {
 
 // New returns a default Config overridden with given parameters.
 func New(uri string, requests, concurrency int, requestTimeout, globalTimeout time.Duration) Config {
-	cfg := Config{
+	var urlURL *url.URL
+	if uri != "" {
+		// ignore err: a Config can be invalid at this point
+		urlURL, _ = url.Parse(uri)
+	}
+	return Config{
 		Request: Request{
+			URL:     urlURL,
 			Timeout: requestTimeout,
 		},
 		RunnerOptions: RunnerOptions{
@@ -44,9 +50,6 @@ func New(uri string, requests, concurrency int, requestTimeout, globalTimeout ti
 			GlobalTimeout: globalTimeout,
 		},
 	}
-	cfg.Request.URL, _ = url.Parse(uri) // TODO: error handling
-
-	return MergeDefault(cfg)
 }
 
 // Default returns a default config that is safe to use.
--- a/config/config.go
+++ b/config/config.go
@@ -7,12 +7,14 @@ import (
 	"time"
 )
 
+// Request contains the confing options relative to a single request.
 type Request struct {
 	Method  string
 	URL     *url.URL
 	Timeout time.Duration
 }
 
+// RunnerOptions contains options relative to the runner.
 type RunnerOptions struct {
 	Requests      int
 	Concurrency   int
@@ -20,6 +22,7 @@ type RunnerOptions struct {
 }
 
 // Config represents the configuration of the runner.
+// It must be validated using Config.Validate before usage.
 type Config struct {
 	Request       Request
 	RunnerOptions RunnerOptions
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -34,8 +34,14 @@ func (cfg Config) String() string {
 
 // New returns a default Config overridden with given parameters.
```
moreirathomas: We may want to update the doc to state that this returns something that may not be valid/usable and that its completely fine as it is meant to be merged before anything else.

config/config.go
```diff
@@ -32,10 +32,16 @@ func (cfg Config) String() string {
 	return string(b)
 }
 
-// New returns a default Config overridden with given parameters.
+// New returns a Config initialized with given parameters.
```
moreirathomas, line 35: Nitpick, we may want to document that is may be invalid/unusable until merged with `config.Merge`.

moreirathomas (approved):
Solves the problem 👌
Just one small nitpick, do as you wish regarding it.

moreirathomas (approved):

PR #26, opened 2022-01-31
dev : implement proper error handling for user input validation

c1-ra:
## Description

Add validation of the user input into the runner CLI in `Config.Validate()` method and related tests.

### DIFF
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -48,6 +48,12 @@ func main() {
 	cfg := makeRunnerConfig()
 	fmt.Println(cfg)
 
+	cfg, err := cfg.Validate()
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
 	if err := requester.New(cfg).RunAndReport(serverURL); err != nil {
 		log.Fatal(err)
 	}
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -50,8 +50,7 @@ func main() {
 
 	cfg, err := cfg.Validate()
 	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
+		log.Fatal(err)
 	}
 
 	if err := requester.New(cfg).RunAndReport(serverURL); err != nil {
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -48,8 +48,7 @@ func main() {
 	cfg := makeRunnerConfig()
 	fmt.Println(cfg)
 
-	cfg, err := cfg.Validate()
-	if err != nil {
+	if err := cfg.Validate(); err != nil {
 		log.Fatal(err)
 	}
 
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -3,7 +3,6 @@ package main
 import (
 	"errors"
 	"flag"
-	"fmt"
 	"log"
 	"os"
 	"time"
@@ -45,10 +44,8 @@ func parseArgs() {
 func main() {
 	parseArgs()
 
-	cfg := makeRunnerConfig()
-	fmt.Println(cfg)
-
-	if err := cfg.Validate(); err != nil {
+	cfg, err := makeRunnerConfig()
+	if err != nil {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -68,6 +68,37 @@ func New(uri string, requests, concurrency int, requestTimeout, globalTimeout ti
...
+// Validate returns the config and a not nil ErrInvalid if any of the fields provided by the user is not valid
+func (cfg Config) Validate() (Config, error) {
+	inputErrors := []error{}
+
+	_, err := url.ParseRequestURI(cfg.Request.URL.String())
+	if err != nil {
+		inputErrors = append(inputErrors, fmt.Errorf("-url: %s is not a valid url", cfg.Request.URL.String()))
+	}
+
+	if cfg.RunnerOptions.Requests < 0 {
```
GregoryAlbouy, line 80: Can we use the new rules for numbers now (`n == -1 || n > 0`)?

Note: if we do, we'll need to update `semimpl.Do`, so we probably should do the error handling for it in the mean time

config/config.go
```diff
@@ -68,6 +68,37 @@ func New(uri string, requests, concurrency int, requestTimeout, globalTimeout ti
...
+		inputErrors = append(inputErrors, fmt.Errorf("-requests: must be >= 0, we got %d", cfg.RunnerOptions.Requests))
+	}
+
+	if cfg.RunnerOptions.Concurrency <= 0 {
```
GregoryAlbouy, line 84: Same here

cmd/runner/main.go
```diff
@@ -48,6 +48,11 @@ func main() {
 	cfg := makeRunnerConfig()
 	fmt.Println(cfg)
 
+	cfg, err := cfg.Validate()
```
GregoryAlbouy, line 51: Actually we don't need to retrieve a new config from `Validate`, we could just do:
```go
if err := cfg.Validate(); err != nil { ... }
```

config/config.go
c1-ra: Let's just do it now!

cmd/runner/main.go
c1-ra: You are right, it is not supposed to change anything in the configuration so it does not need to return it. I have changed it!

config/config.go
```diff
@@ -68,6 +68,41 @@ func New(uri string, requests, concurrency int, requestTimeout, globalTimeout ti
...
+func (cfg Config) Validate() error {
+	inputErrors := []error{}
+
+	_, err := url.ParseRequestURI(cfg.Request.URL.String())
+	if err != nil {
+		inputErrors = append(inputErrors, fmt.Errorf("-url: %s is not a valid url", cfg.Request.URL.String()))
+	}
+
+	if cfg.RunnerOptions.Requests < 0 {
+		if cfg.RunnerOptions.Requests != -1 {
```
GregoryAlbouy, line 81: We can avoid that extra nesting using `&&`, also I think `0` should be an invalid values for `requests`:
```suggestion
	if cfg.RunnerOptions.Requests < 1 && cfg.RunnerOptions.Requests != -1 {
```

config/config.go
```diff
@@ -68,6 +68,41 @@ func New(uri string, requests, concurrency int, requestTimeout, globalTimeout ti
...
+			inputErrors = append(inputErrors, fmt.Errorf("-requests: must be >= 0, we got %d", cfg.RunnerOptions.Requests))
+		}
+	}
+
+	if cfg.RunnerOptions.Concurrency < 0 {
+		if cfg.RunnerOptions.Concurrency != -1 {
```
GregoryAlbouy, line 87: Actually I realize we probably shouln't accept any value below `1` for `concurrency`, because `1` already stands for "no concurrency" 🤔

GregoryAlbouy (commented):
Nice! Just a couple of changes for me and we're good to go!

config/config.go
c1-ra: I first used `&&`, but `gocognit` was complaining that it was too complex :(
It is right though, 0 should also be invalid.

config/config.go
c1-ra: Indeed!

config/config.go
GregoryAlbouy: Well if `gocognit` considers a single `&&` more complex than a nested `if` then it doesn't do its job correctly 🤷‍♂️ 
I suggest we shut it down and remove it from `.golangci.yml`.

c1-ra: I have used `//nolint `on the method for the moment, if we run into this problem again, we can remove `gocognit` completely...

GregoryAlbouy (approved):

moreirathomas (approved):

PR #28, opened 2022-02-01
fix: zero-values cli flags do not override config file

GregoryAlbouy:
## Description

Due to the behaviour of `config.Merge` that relies on zero values to decide whether an option should be overridden, it is currently not possible to override a config file option with a CLI option that is a zero value.
For now, it's not that much of an issue since any zero value is an invalid value. However this is not guaranteed to last: in the future we can implement options accepting `0` as a valid value (such as `interval`) or bools, for which obviously `false` should be acceptable.

So this PR provides a new overriding logic via `Config.Override` based on which values are actually set rather than zero values.

## Changes

- Replace `config.Merge`, `config.MergeDefault` with `Config.Override` that allows to replace fields selectively
- Fix override baseConfig ⬅️ cliConfig: set fields are retrieved via `flag.CommandLine.Visit`
- Fix override baseConfig ⬅️ fileConfig: set fields are retrieved via non-nil pointer types in `config/file.marshaledConfig`


cmd/runner/main.go
```diff
@@ -57,36 +56,22 @@ func main() {
...
 
-	return cfg
+	return fileCfg.Override(cliCfg, flagNames()...)
 }
 
-// overrideConfigWithCLIFlags replaces config.Config values for every
-// corresponding CLI flag set. It must be called after flag.Parse.
-func overrideConfigWithCLIFlags(cfg *config.Config) {
+// flagNames returns a slice of all flags set.
+func flagNames() []string {
```
c1-ra, line 71: If find the use of this `flagNames()` method more elegant than the switch! 👍

config/config.go
```diff
@@ -46,6 +46,28 @@ func (cfg Config) HTTPRequest() (*http.Request, error) {
...
+			cfg.RunnerOptions.Requests = c.RunnerOptions.Requests
+		case "concurrency":
+			cfg.RunnerOptions.Concurrency = c.RunnerOptions.Concurrency
+		case "globalTimeout":
+			cfg.RunnerOptions.GlobalTimeout = c.RunnerOptions.GlobalTimeout
+		}
+	}
+	return cfg
+}
+
```
c1-ra, line 70: I think it makes sense that it is a method of `Config` and not in `main`. But I have no strong opinion about this, so the first implementation works for me too 🤷‍♀️

c1-ra:
I have added some comments, but both versions could work as they are for me!

moreirathomas:
Solves the problem but the api becomes a bit confused:
- we can call `Merge` from main but it is not intended to be used this way
- `Merge` is only used through `MergeDefault`
- `Config.Merge` can be called from `requester` when we said that importing package `config` should not come with a parser (overriding is kinda concerned with parsing...)

Suggestion:
Ideally `config` exposes only `Config` struct and `Config.HTTPRequest()`.
We could then have something like `config/parse` with `parse.FromFile()`, `parse.FromFlags()`, `parse.Override(base, override config.Config)`.

Or this could not work, but anyway I think the current API is not really what we want.
We may put this on hold until we have something better and simply merge the simplest fix to the bug ?

GregoryAlbouy:
Some clarification about the PR (my bad for not pointing that up earlier):

- It's not ready to merge as is: the same fix has to be done for the config file defaulting: because it still uses `Merge` (with zero values logics), we currently can't override a `DefaultConfig` option with a zero-value from the config file

- In other words `Merge`/`MergeDefault` **will be removed**: their only usage is from `config/file.parseRawConfig`, which has to be replaced with `Config.Override` logics.

> - `Config.Merge` can be called from `requester` when we said that importing package `config` should not come with a parser (overriding is kinda concerned with parsing...)

I disagree with that point: there's factually no parsing involved in `Merge` or `Override`, only values that are copied or not from one Config to another under certain conditions. They just return a `config.Config` out of `config.Config`s, hence they should remain under `config`.

> We could then have something like config/parse with `parse.FromFile()`, `parse.FromFlags()`

That's interesting, but indeed not doable in a way that makes sense: all flags are already parsed before we get to call `parse.FromFlags`, so it becomes `parse.FromAlreadyParsedFlags`, which is nothing different from `config.New`.

config/file/parser.go
```diff
@@ -33,3 +33,22 @@ func newParser(ext extension) (configParser, error) {
...
+		QueryParams map[string]string `yaml:"queryParams" json:"queryParams"`
+		Timeout     *string           `yaml:"timeout" json:"timeout"`
+	} `yaml:"request" json:"request"`
+
+	RunnerOptions struct {
+		Requests      *int    `yaml:"requests" json:"requests"`
+		Concurrency   *int    `yaml:"concurrency" json:"concurrency"`
+		GlobalTimeout *string `yaml:"globalTimeout" json:"globalTimeout"`
+	} `yaml:"runnerOptions" json:"runnerOptions"`
+}
```
moreirathomas, line 54: I see by the commit message that somehow using pointers to type fixes zero-value override.
As I did not see the bug happening, do you care to explain a bit ? :)

GregoryAlbouy: Sure! In short, it's only a necessary step to make `parseRawConfig` where lies the fix:
https://github.com/benchttp/runner/blob/22cb9d071a839c652e6300b96bb8715a69516acd/config/file/parse.go#L35-L86

Just like for CLI flags with `main.configFlags`, we need to know which options are set regardless of their value. But this is much more complex to achieve for parsed json/yaml, so here we use pointers instead to determine whether a field is set. It's handy because their zero value is `nil`: without them it would be `""` or `0` and we wouldn't be able to tell if it's defaulted or an actual value from the config file.

c1-ra (approved):
Nice! It looks good to me!

c1-ra:
Maybe you could just re-add the [comment that Thomas suggested](https://github.com/benchttp/runner/pull/28#discussion_r798917385) concerning `Requests = -1` ?

config/file/parser.go
moreirathomas: Yeah obviously ! relying on nil is a great way of doing this !

config/default.go
```diff
@@ -13,7 +13,7 @@ var defaultConfig = Config{
 	},
 	RunnerOptions: RunnerOptions{
 		Concurrency:   1,
-		Requests:      -1,
+		Requests:      -1, // Use GlobalTimeout as exit condition if omitted.
```
moreirathomas, line 16: Nitpick-but-not-a-nitpick: explicit `-1` results is the use of global timeout as exit condition. Omitting the options results in defaulting to `-1` thus the global timeout.

GregoryAlbouy: Isn't it already implied given this comment is on `defaultConfig`?
On a side note we really need to rethink our defaults, I don't think infinite requests by default is a great idea, especially with a default timeout of `30s` 🤔  `100` would be more appropriate

c1-ra: I agree that a limited default value would actually be safer.

c1-ra (dismissed):

moreirathomas (approved):

c1-ra (approved):

PR #29, opened 2022-02-01
feat: implement proper error handling for requester

GregoryAlbouy:
## Description

This PR implements proper error handling for `requester` and `dispatcher` (formerly `semimpl`).

### Refactoring: `semimpl` ➡️ `dispatcher`
  ```go
  // before
  semimpl.Do(ctx, 100, 10, callback) // notice the error?
  
  // after
  dispatcher.New(10).Do(ctx, 100, callback)
  ```

### Error handling: `dispatcher`

- `dispatcher.New` **panics** if `numWorker < 1`
- `dispatcher.Dispatcher.Do` returns a non-nil error early if
  - `maxIter < 1 && maxIter != -1`
  - `maxIter < Dispatcher.numWorker && maxIter != -1`
  - `callback == nil`
- New unit tests for all cases

### Error handling: `requester`
- Handle `config.HTTPRequest()` error in `Dispatcher.Run` instead of repeated callback
- Handle `client.Do` connection error in `Dispatcher.Run` in addition to the callback via a ping to return early in case of bad URL, server not started, ...

### DIFF
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -1,4 +1,4 @@
-package semimpl
+package dispatcher
 
 import (
 	"context"
@@ -7,21 +7,23 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
+type Dispatcher struct {
+	sem *semaphore.Weighted
+}
+
 // Do concurrently executes callback at most maxIter times or until ctx is done
 // or canceled. Concurrency is handled leveraging the semaphore pattern, which
-// ensures at most numWorkers goroutines are spawned at the same time.
-func Do(ctx context.Context, numWorkers, maxIter int, callback func()) {
-	numWorkers = sanitizeNumWorkers(numWorkers)
+// ensures at most Dispatcher.numWorkers goroutines are spawned at the same time.
+func (d Dispatcher) Do(ctx context.Context, maxIter int, callback func()) {
 	maxIter = sanitizeMaxIter(maxIter)
 	callback = sanitizeCallback(callback)
 
-	sem := semaphore.NewWeighted(int64(numWorkers))
 	wg := sync.WaitGroup{}
 
 	for i := 0; i < maxIter || maxIter == 0; i++ {
 		wg.Add(1)
 
-		if err := sem.Acquire(ctx, 1); err != nil {
+		if err := d.sem.Acquire(ctx, 1); err != nil {
 			// err is either context.DeadlineExceeded or context.Canceled
 			// which are expected values so we stop the process silently.
 			wg.Done()
@@ -30,7 +32,7 @@ func Do(ctx context.Context, numWorkers, maxIter int, callback func()) {
 
 		go func() {
 			defer func() {
-				sem.Release(1)
+				d.sem.Release(1)
 				wg.Done()
 			}()
 			callback()
@@ -40,7 +42,14 @@ func Do(ctx context.Context, numWorkers, maxIter int, callback func()) {
 	wg.Wait()
 }
 
-func sanitizeNumWorkers(numWorkers int) int {
+// New returns a Dispatcher initialized with numWorker.
+func New(numWorker int) Dispatcher {
+	numWorker = sanitizeNumWorker(numWorker)
+	sem := semaphore.NewWeighted(int64(numWorker))
+	return Dispatcher{sem: sem}
+}
+
+func sanitizeNumWorker(numWorkers int) int {
 	if numWorkers < 1 {
 		return 1
 	}
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -30,7 +30,7 @@ func (d dispatcher) Do(ctx context.Context, maxIter int, callback func()) error 
 
 	wg := sync.WaitGroup{}
 
-	for i := 0; i < maxIter || maxIter == 0; i++ {
+	for i := 0; i < maxIter || maxIter == -1; i++ {
 		wg.Add(1)
 
 		if err := d.sem.Acquire(ctx, 1); err != nil {
@@ -54,10 +54,10 @@ func (d dispatcher) Do(ctx context.Context, maxIter int, callback func()) error 
 }
 
 func (d dispatcher) validate(maxIter int, callback func()) error {
-	if maxIter < 1 {
-		return fmt.Errorf("%w: maxIter: must be < 1, got %d", ErrInvalidValue, maxIter)
+	if maxIter < 1 && maxIter != -1 {
+		return fmt.Errorf("%w: maxIter: must be -1 or >= 1, got %d", ErrInvalidValue, maxIter)
 	}
-	if maxIter < d.numWorker {
+	if maxIter < d.numWorker && maxIter != -1 {
 		return fmt.Errorf(
 			"%w: maxIter: must be >= numWorker (%d), got %d",
 			ErrInvalidValue, d.numWorker, maxIter,
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -2,6 +2,7 @@ package dispatcher
 
 import (
 	"context"
+	"fmt"
 	"sync"
 
 	"golang.org/x/sync/semaphore"
@@ -48,16 +49,11 @@ func (d dispatcher) Do(ctx context.Context, maxIter int, callback func()) {
 
 // New returns a Dispatcher initialized with numWorker.
 func New(numWorker int) Dispatcher {
-	numWorker = sanitizeNumWorker(numWorker)
+	if numWorker < 1 {
+		panic(fmt.Sprintf("invalid numWorker value: must be > 1, got %d", numWorker))
+	}
 	sem := semaphore.NewWeighted(int64(numWorker))
 	return dispatcher{sem: sem}
-}
-
-func sanitizeNumWorker(numWorkers int) int {
-	if numWorkers < 1 {
-		return 1
-	}
-	return numWorkers
 }
 
 func sanitizeMaxIter(maxIter int) int {
--- a/dispatcher/dispatcher.go
+++ b/dispatcher/dispatcher.go
@@ -7,14 +7,18 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
-type Dispatcher struct {
+type Dispatcher interface {
+	Do(ctx context.Context, maxIter int, callback func())
+}
+
+type dispatcher struct {
 	sem *semaphore.Weighted
 }
 
 // Do concurrently executes callback at most maxIter times or until ctx is done
 // or canceled. Concurrency is handled leveraging the semaphore pattern, which
 // ensures at most Dispatcher.numWorkers goroutines are spawned at the same time.
-func (d Dispatcher) Do(ctx context.Context, maxIter int, callback func()) {
+func (d dispatcher) Do(ctx context.Context, maxIter int, callback func()) {
 	maxIter = sanitizeMaxIter(maxIter)
 	callback = sanitizeCallback(callback)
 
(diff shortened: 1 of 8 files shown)


c1-ra (dismissed):
[question] Why do you think this way is less prone to error?
I am ok you merge it anyway, I have really no preference between this implementation and the previous one, both seem ok to me!

GregoryAlbouy:
It is less error prone because it avoids this kind of error:
```go
semimpl.Do(ctx, 100, 10, callback) // notice the error?
                                   // numWorker and maxIter are inverted: we have more workers than iterations
```

dispatcher/dispatcher.go
```diff
@@ -40,7 +42,14 @@ func Do(ctx context.Context, numWorkers, maxIter int, callback func()) {
 	wg.Wait()
 }
 
-func sanitizeNumWorkers(numWorkers int) int {
+// New returns a Dispatcher initialized with numWorker.
+func New(numWorker int) Dispatcher {
+	numWorker = sanitizeNumWorker(numWorker)
```
moreirathomas, line 47: At this point why not inline this function ?
```suggestion
	if numWorkers < 1 {
		numWorkers = 1
	}
```

Same for the 2 others?

moreirathomas (commented):
Seems good.

Same as Clara, I don't know if an other data struct is really necessary but
> ```go
> semimpl.Do(ctx, 100, 10, callback) // notice the error?
>                                    // numWorker and maxIter are inverted: we have more workers than iterations
> ```
is valid so I'll run with it.

Just waiting for reply on the other comment before approving.

dispatcher/dispatcher.go
c1-ra: Since we are supposed to call `Validate()` on the config before declaring the dispatcher, couldn't we just get rid of all the sanitize methods because we can suppose that all configurations are valid?

GregoryAlbouy:
>  I don't know if an other data struct is really necessary

Clearly none of the changes in this PR is necessary. Only debatable changes from `semimpl` that I consider an API improvement.

One notably questionable consequence is regarding the error handling, we probably don't want to check both `New` and `Do`:
```go
d, err := dispatcher.New(-4)
if err != nil {}
if err = d.Do(...) {}
```
Because this si clearly not an AP improvement. We'll want to keep it chainable `dispatcher.New(10).Do(...)` so we'll have to silently sanitize `numWorkers` in `New` and do the error handling in `Do` only (which is perfectly ok to me)

dispatcher/dispatcher.go
GregoryAlbouy: To me `dispatcher` should remain a completely independent package with its own logic, and making assumptions on its inputs based on how it's used outside is already a form of couplage. The external config validation is subject to change, and this package shouldn't depend on it. For that reason I believe it should have its own error handling.

As for the inlining, agreed!

dispatcher/dispatcher_test.go
```diff
@@ -244,50 +245,51 @@ func TestDo(t *testing.T) {
...
+	testcases := []struct {
+		label     string
+		exp       error
+		numWorker int
+		maxIter   int
+		callback  func()
+	}{
+		{
+			label:     "return error if maxIter < 1",
+			exp:       errors.New("invalid value: maxIter: must be < 1, got 0"),
```
c1-ra, line 257: `exp: errors.New("invalid value: maxIter: must be >= 1, got 0")`
I think it should be **>=1**?

dispatcher/dispatcher.go
```diff
@@ -0,0 +1,79 @@
...
+		}()
+	}
+
+	wg.Wait()
+	return nil
+}
+
+func (d dispatcher) validate(maxIter int, callback func()) error {
+	if maxIter < 1 {
+		return fmt.Errorf("%w: maxIter: must be < 1, got %d", ErrInvalidValue, maxIter)
```
c1-ra, line 58: `return fmt.Errorf("%w: maxIter: must be >= 1, got %d", ErrInvalidValue, maxIter)`

c1-ra:
I just added two minor comments, otherwise, it seems good :)

dispatcher/dispatcher.go
GregoryAlbouy: Good catch! Not minor at all, totally forgot that use case. Fixed!

requester/requester.go
```diff
@@ -19,7 +26,9 @@ type Requester struct {
 	tracer *tracer
 }
 
-// New returns a Requester configured with specified Options.
+// New returns a Requester initialized with cfg. cfg is assumed valid:
+// it is the caller's responsibility to ensure cfg is valid using
+// cfg.Validate.
 func New(cfg config.Config) *Requester {
 	tracer := newTracer()
 	return &Requester{
```
moreirathomas, line 34: Side not but super important: now that `-1` for number of requests is allowed and expected, calling `make(chan Record, cfg.RunnerOptions.Requests)` can panic because `make` do not accept negative integers.

requester/requester.go
```diff
@@ -40,20 +49,41 @@ func New(cfg config.Config) *Requester {
...
-func (r *Requester) Run() Report {
+func (r *Requester) Run() (Report, error) {
+	req, err := r.config.HTTPRequest()
+	if err != nil {
+		return Report{}, fmt.Errorf("%w: %s", ErrRequest, err)
+	}
+
+	if err := r.ping(req); err != nil {
+		return Report{}, fmt.Errorf("%w: %s", ErrConnection, err)
+	}
```
moreirathomas, line 60: Very nice !

requester/requester.go
```diff
@@ -40,20 +49,41 @@ func New(cfg config.Config) *Requester {
...
+			New(r.config.RunnerOptions.Concurrency).
+			Do(ctx, r.config.RunnerOptions.Requests, r.record(req))
 	}()
 
-	return r.collect()
+	if err := <-errCh; err != nil {
+		return Report{}, err
+	}
+
+	return r.collect(), nil
```
moreirathomas, line 78: I might be wrong but:
I'm pretty sure reading from `errCh` will block until something from `requester.Do` writes to it.

Meaning `r.collect()` is not run in parallel of `Do`. Losing the whole point.

---

Suggestion/path to explore, untested:

```suggestion
	return r.collect(ctx, errCh)
```
And this `r.collect`:
```go
func (r *Requester) collect(ctx context.Context, errc <-chan error) (Report, error) {
	rep := Report{
		Config:  r.config,
		Records: make([]Record, 0, r.config.RunnerOptions.Requests), // Provide capacity if known.
	}
	for {
		select {
		case <-ctx.Done():
			return rep, nil
		case err := <-errc:
			return Report{}, err
		case rec := <-r.records:
			if rec.Error != nil {
				rep.Fail++
			} else {
				rep.Success++
			}
			rep.Records = append(rep.Records, rec)
			rep.Length++
		}
	}
}
```

Note: we go back to incremental Report building.

Maybe going back to safe slice and mutex is better.

requester/requester.go
GregoryAlbouy: Indeed, noticed the same for `collect`:
https://github.com/benchttp/runner/blob/300bfb6438467cfd86ea38bf58871a6e423e7a6f/requester/report.go#L30-L34

Hence the addition in the docs 3 lines above this comment:
```go
// New returns a Requester initialized with cfg. cfg is assumed valid:
// it is the caller's responsibility to ensure cfg is valid using
// cfg.Validate.
func New(cfg config.Config) *Requester {
```

requester/requester.go
GregoryAlbouy: Oh nice catch indeed. Will fix it as you suggest, and propose a second PR with mutexes targeting this one so we can compare.

requester/requester.go
moreirathomas: But -1 _is_ a valid config isn't it ? Or did I miss a change ?

Thus we can pass -1 in `make`.

GregoryAlbouy: Damn why do I keep forgetting that.

requester/report.go
```diff
@@ -27,43 +31,52 @@ func (rep Report) String() string {
...
 		Config:  r.config,
-		Records: make([]Record, 0, r.config.RunnerOptions.Requests), // Provide capacity if known.
+		Records: make([]Record, 0, recordsCap), // Provide capacity if known.
 	}
 
-	for rec := range r.records {
+	for rec := range r.recordC {
+		select {
+		case err := <-r.errC:
+			return Report{}, err
```
moreirathomas, line 48: Why not call `cancel` here ? We stopped the process by the context with duration countdown is still going, and `errC` is not closed either.

Maybe we should not handle this here, but I think we should handle it anyway.
So either pass `cancel` to `collect` or read error inside `Run` and cancel before returning.

requester/report.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/benchttp/runner/config"
 )
 
+const (
+	defaultRecordsCap = 128
```
moreirathomas, line 13: Why 128 specifically ? Does it get reallocated if it exceeds?

requester/report.go
GregoryAlbouy: `cancel` is already called in `Run`. It has nothing to do with error handling: we use `cancel` to close it once we're done we it to avoid leaks (similarly to `io.ReadCloser`s `Close`), so the right place to do it is just after calling `Dispatcher.Do(ctx, ...)`, regardless of how we handle the error it returns.

requester/report.go
GregoryAlbouy: Because why not 😛 The cap algorithm is very simple: it reallocates the double when reached. So for 1000 requests, it'll be reallocated 3 times.
These amount of space are not gonna be an issue, so I think we can go for `1000` by default.

requester/report.go
```diff
@@ -23,60 +23,53 @@ func (rep Report) String() string {
...
 
-// collect pulls the records from Requester.Records as soon as
-// they are available and consumes them to build the report.
-// Returns the report when all the records have been collected.
-// Requester.collect will blocks until Requester.Records is empty.
-func (r *Requester) collect() Report {
-	rep := Report{
+// report generates and returns a Report from a previous Run.
+func (r *Requester) report() Report {
+	return Report{
```
moreirathomas, line 28: Should not happen but should we noop if `Run` was not called, ie `len(r.records) == 0 && r.numErr == 0` ?

moreirathomas, line 28: Should not happen but should we noop if `Run` was not called, ie `len(r.records) == 0 && r.numErr == 0` ?

moreirathomas (approved):
Small question of small incidence. I'll approve right now, but if you think you need to address it, i'll re-approve again :)

requester/report.go
GregoryAlbouy: I wouldn't bother given it's private + it would return a zero Report anyway (except for the config).
We can consider making it a package func as a foolproof:
```go
func makeReport(cfg config.Config, records []Record, numErr int) Report
```

moreirathomas: Let's put that as a "nice to have" and move on then :)

GregoryAlbouy: I'm fine making the change if we judge it preferable, it's a matter of seconds

dispatcher/dispatcher_test.go
```diff
@@ -31,24 +66,25 @@ func TestDo(t *testing.T) {
...
+			maxIter   = -1
 
 			margin      = 25 * time.Millisecond // determined empirically
 			maxDuration = timeout + margin
 		)
 
 		var (
-			maxIter = int(interval.Milliseconds()) + 1 // should not be reached
-			gotIter = 0
+			expIterMax = int(interval.Milliseconds()) + 1 // should not be reached
```
c1-ra, line 79: [question]
Is `expIterMax` determined empirically too?

requester/report.go
c1-ra: If it is quick, I think it does not eat bread :bread: to add it. I guess it makes more explicit that we have thought about this possibility and that it will not return a report in this case.

c1-ra (approved):
:+1:

moreirathomas (approved):

c1-ra (approved):
Nice!

PR #35, opened 2022-02-03
dev: refactor config parsing

GregoryAlbouy:
## Description

Exploration following [that comment](https://github.com/benchttp/runner/pull/28#issuecomment-1027730797).
The main idea is to move all the input config parsing into a single package, including the CLI flags parsing.
If we find that base acceptable (we may not), let's iterate on it!

## Changes

- Set flags values directly in a `config.Global` struct
- Move package `config/file` ➡️ `internal/configfile`
- Create package `internal/configflags` for CLI flags parsing implementation
- Move runner config flags under new subcommand `run`
  - avoids settings flags globally in external package
  - prepares further subcommands such as `benchttp auth`
  - usage:
    ```sh
    # before
    benchttp -url "http://localhost:9999"
    
    # after
    benchttp run -url "http://localhost:9999"
    ```
- Write unit tests for `internal/configflags`
- Namespace subcommand methods in structs
  - `run` ➡️  `cmdRun` (`cmd/benchttp/run.go`)
  - `auth` ➡️  `cmdAuth` (`cmd/benchttp/auth.go`)


GregoryAlbouy:
Not 100% convinced about moving the config CLI flags parsing into `config/parse` 😕 
Having `parse.Flags` and `parse.File` may sound nice, but when we add new command such as `login` or else, we'll want this specific logic gathered in the same place rather than scattered `config/parse`, `main` (or else).

c1-ra (approved):
Great work, it was a lot of refactoring!

GregoryAlbouy:
Conflict after merge of #72, automatically resolved using `git rebase --onto main fix/use-new-endpoint`.

cmd/benchttp/run.go
```diff
@@ -0,0 +1,151 @@
...
+		return
+	}
+
+	mergedConfig := fileConfig.Override(cmd.config, fields...)
+
+	return mergedConfig, mergedConfig.Validate()
+}
+
+// requesterConfig returns a requester.Config generated from cfg.
+func (cmd cmdRun) requesterConfig(cfg config.Global) requester.Config {
```
moreirathomas, line 115: Nitpick: it seems it does not need to be a method of `cmdRun`.

I understand it is to limit the access to this to `cmdRun`, but is it a good enough reason ? Really question, I'd like you're view on that.

cmd/benchttp/run.go
```diff
@@ -0,0 +1,151 @@
...
+		RequestTimeout: cfg.Runner.RequestTimeout,
+		GlobalTimeout:  cfg.Runner.GlobalTimeout,
+		Silent:         cfg.Output.Silent,
+	}
+}
+
+// listenOSInterrupt listens for OS interrupt signals and calls callback.
+// It should be called in a separate goroutine from main as it blocks
+// the execution until the OS interrupt signal is received.
+func (cmd cmdRun) listenOSInterrupt(callback func()) {
```
moreirathomas, line 129: Nitpick: it seems it does not need to be a method of cmdRun either. Same question.

cmd/benchttp/run.go
```diff
@@ -0,0 +1,151 @@
...
+	sigC := make(chan os.Signal, 1)
+	signal.Notify(sigC, os.Interrupt)
+	<-sigC
+	callback()
+}
+
+// handleRunInterrupt handles the case when the runner is interrupted.
+func (cmd cmdRun) handleRunInterrupt() error {
```
moreirathomas, line 137: Nitpick: it seems it does not need to be a method of cmdRun still. Same question :)

config/field.go
```diff
@@ -15,12 +15,23 @@ const (
 	FieldTemplate       = "template"
 )
 
+// FieldsDesc is a record of all available config fields and their description.
+var FieldsDesc = map[string]string{
```
moreirathomas, line 19: Suggestion: even with the documentation, I feel like `Desc` as too much meaning as `descending` (`asc` vs `desc`).

`FieldsDefinition`, `FieldsUsage` ?

internal/configflags/set.go
```diff
@@ -0,0 +1,92 @@
...
+	"flag"
+	"net/http"
+	"net/url"
+
+	"github.com/benchttp/runner/config"
+)
+
+// Set reads arguments provided to flagset as config.Fields and binds
+// their value to the appropriate fields of given *config.Global.
+func Set(flagset *flag.FlagSet, cfg *config.Global) {
```
moreirathomas, line 13:
```suggestion
// Set reads arguments provided to flagset as config.Fields and binds
// their value to the appropriate fields of given *config.Global.
// The provided *flag.Flagset must not have been parsed yet, otherwise
// bindings its values would fail.
func Set(flagset *flag.FlagSet, dst *config.Global) {
```

I think var name `dst` would convey more meaning about what's going on.

internal/configflags/which.go
```diff
@@ -0,0 +1,19 @@
+package configflags
+
+import (
+	"flag"
+
+	"github.com/benchttp/runner/config"
+)
+
+// Which returns a slice of all config fields set via the CLI
+// for the given FlagSet
```
moreirathomas, line 10:
```suggestion
// for the given *flag.FlagSet.
```

moreirathomas (commented):
Very very nice !

`configflags` is a much cleaner approach than the previous "dump all in main", really like it!

Also setups the organization for other commands.

Some suggestion regarding naming and docs.

Only real thing I'd like we address is the `cmdRun` literal in `main`, more details in the review comments.

cmd/benchttp/run.go
GregoryAlbouy: Well 2 reasons:
- namespacing indeed, to avoid polluting the package scope that is expected to manage very different things
- its i/o `config.Global` and `requester.Config` are 100% specific to the `run` command (they have nothing to do with with other commands)

So to me it makes perfect sense as a method of `cmdRun`. Could also be a function of some package `internal/adapters` but we stated it's bit overkill for now. I don't mind going this way though, otherwise I'd rather keep the package as clean as possible.

Side question: what do you mean exactly by _it does not need_? It seems to imply that it comes at some sort of cost compared to package-scoped functions, what would that be?
If it's only the extra copy of `cmdRun` we can still use a pointer receiver instead. But note that the structure is 208 bytes (we set the linter `hugeParam` to complain at 256) and this method is not expected to be called more than once in the whole program, so I think we're fine with that.

cmd/benchttp/run.go
GregoryAlbouy: Well same answer as https://github.com/benchttp/runner/pull/35#discussion_r815693485: what it does is specific to the `run` command. It's so specific that it could be declared/inlined inside `cmd.Execute`, but I extracted it for better readability. However I don't see the point of extracting it one step further at the package scope 🤔

cmd/benchttp/main.go
```diff
@@ -0,0 +1,53 @@
...
+	var cmd command
+	args := os.Args[1:]
+
+	switch sub := args[0]; sub {
+	case "run":
+		cmd = cmdRun{
+			flagset:            flag.NewFlagSet("run", flag.ExitOnError),
+			config:             config.Default(),
+			defaultConfigFiles: defaultConfigFiles,
+		}
```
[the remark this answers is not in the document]
GregoryAlbouy: Well technically we could even declare `flagset` inside `cmdRun`, it would work equally.
Here it's just about injecting `cmdRun` dependencies rather than relying on globals so it remains predictable based on its input and more easily testable. It also has the advantage to avoid the need of a `newRunCommand` constructor.

So to me it's far from being the obvious way to go, although I agree `cmdRun` defining its own defaults makes perfect sense, so it's clearly worth considering.

cmd/benchttp/run.go
```diff
@@ -0,0 +1,151 @@
...
+	"os/signal"
+
+	"github.com/benchttp/runner/config"
+	"github.com/benchttp/runner/internal/configfile"
+	"github.com/benchttp/runner/internal/configflags"
+	"github.com/benchttp/runner/output"
+	"github.com/benchttp/runner/requester"
+)
+
+var defaultConfigFiles = []string{
```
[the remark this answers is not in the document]
GregoryAlbouy: See https://github.com/benchttp/runner/pull/35#discussion_r815689476.

Your description is deliberately convoluted, but in better words we simply inject a value in an independent structure. That value happens to be available in the global scope: it could not be as well, or change at runtime without affecting `cmdRun`.

internal/configflags/set.go
GregoryAlbouy: Yup!
As for the doc addition, aren't we (if not stating the obvious) trying to explain how Go generally works regarding flags? The change is OK to me in any case.

cmd/benchttp/main.go
```diff
@@ -0,0 +1,53 @@
...
+	if err := run(); err != nil {
+		fmt.Println(err)
+		if errors.Is(err, errUsage) {
+			flag.Usage()
+		}
+		os.Exit(1)
+	}
+}
+
+func run() error {
```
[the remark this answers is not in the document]
GregoryAlbouy: `start`? `bootstrap`?

Bonus non-suggestion:
realwrappedmain

cmd/benchttp/run.go
moreirathomas: > Side question: what do you mean exactly by it does not need?

Because it really does not _need_ the method receiver `(cmd cmdRun)`.
Namespacing this way is valid too. I'm totally fine with doing this.

cmd/benchttp/main.go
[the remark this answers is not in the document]
moreirathomas: Fair enough !

I have a slight preference for a `newRunCommand` returning a ready command, without relying on values injection (ie inlining `config.Default()` and `defaultConfigFiles` inside the constructor).

But both are virtually equal, do what you think is best. The suggestion was worth sharing :)

cmd/benchttp/main.go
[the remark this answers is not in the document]
moreirathomas: As I said, I have no idea :laughing:
If you find something meaningful go ahead, otherwise its fine !

cmd/benchttp/main.go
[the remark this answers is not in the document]
GregoryAlbouy: What do you think of 0d50cbdfc9849508d93f49da66578fb08fb60975`Set cmdRun defaults in method init`?

It's a bit of a compromise:
- defaults are set internally, so we lose a bit of visibility about our defaults but it clears the package
- no `new` constructor but an `init` method called in `execute`: we lose in terms of testability but no new declaration in the package

Fine for me. Better or worse for you?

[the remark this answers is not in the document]
moreirathomas: Seems good to me too ! If we go this way, I would go down the route and make a constructor honestly.
It also removes the "ensure implements interface stuff".

```go
// command is the interface that all benchttp subcommands must implement.
type command interface {
	execute(args []string) error
}

// ensure cmdRun implements command
// var _ command = (*cmdRun)(nil) <- useless

func newRunCommand(fs *flag.FlagSet) command {
	return &cmdRun{
		flagset: fs,
		config:  config.Default(),
		defaultConfigFiles: []string{"./.benchttp.yml", "./.benchttp.yaml", "./.benchttp.json"},
	}
}

func run () {
	var cmd command
	args := os.Args[1:]

	switch sub := args[0]; sub {
	case "run":
		cmd = newRunCommand(flag.NewFlagSet("run", flag.ExitOnError))
	// ...
	}
}
```

> defaults are set internally, so we lose a bit of visibility about our defaults but it clears the package

Well technically true, but I don't think that reading the next struct literal gives us much more info than a constructor: 

```go
cmd = &cmdRun{
	flagset:            flag.NewFlagSet("run", flag.ExitOnError),
	config:             config.Default(),
	defaultConfigFiles: defaultConfigFiles,
}
```

In any case, I do not require any changes or press anything, really go for what you feel is the best.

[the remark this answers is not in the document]
GregoryAlbouy: >  If we go this way, I would go down the route and make a constructor honestly.

To me the constructor is not the logical next step here. I don't get the necessity of it, and it'll add either many unnecessary declarations with commands that need no setup (`help`, `version`...?) or inconsistency in their usage  if we decide not to use constructors for them.
`var _ itf = (*impl)(nil)` has always been anecdotic, we use it for immediate feedback near the implementation, but even without a bad implementation would break at line `cmd = cmdRun{}` 🤷‍♂️ 

> Well technically true, but I don't think that reading the next struct literal gives us much more info than a constructor

Yes, I was referring to the package-level `defaultConfigFiles` at the top of the file 😅

cmd/benchttp/main.go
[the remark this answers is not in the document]
c1-ra: I do not mind keeping `run`, but if we want to change it, `start` is nice too.

c1-ra (approved):

moreirathomas (approved):

PR #42, opened 2022-02-05
ci: unit test config file parsing errors

GregoryAlbouy:
## Description

Original intent was to improve test coverage of package `config/file` by covering the different cases of bad input file (not found, bad extension, parse error, ...)

Doing so eventually led me to improve the error handling of the package by making it more actionnable with defined errors, and more user-friendly by its output messages.

For instance, for an inexistant field set in a yaml file:

| Before | After |
| --- | --- |
| `line 11: field interval not found in type struct { Requests *int "yaml:\"requests\" json:\"requests\""; Concurrency *int "yaml:\"concurrency\" json:\"concurrency\""; GlobalTimeout *string "yaml:\"globalTimeout\" json:\"globalTimeout\"" }`|  `line 11: field "interval" does not exist` |

### DIFF
--- a/config/file/parser.go
+++ b/config/file/parser.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"regexp"
 
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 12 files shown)


config/file/parser.go
```diff
@@ -15,23 +18,167 @@ const (
...
+	// filter out unwanted errors
+	filtered := &yaml.TypeError{}
+	for _, msg := range typeError.Errors {
+		// With decoder.KnownFields set to true, Decode reports any field
+		// that do not match the destination structure as a non-nil error.
+		// It is a wanted behavior but prevents the usage of custom aliases.
+		// To work around this we allow an exception for that rule with fields
+		// starting with x- (inspired by docker compose api).
+		if p.isCustomFieldError(msg) {
+			continue
```
moreirathomas, line 71: Quick question, as I do not know that much about yaml aliases: 
It seems like `decoder.KnownFields(true)` will interpret alias declarations as errors when decoding the raw yaml. But we discard it via `continue`.
I suppose that when decoding the known field refering to the alias the correct value is set, right ?

This means that `KnownFields(true)` is natively incompatible with aliases and this custom implementation is a work-around this limitation ?

moreirathomas (approved):
Nice improvement ! :ok_hand:
Just a side question.

PR #43, opened 2022-02-05
Feat/add config option request body

c1-ra:
## Description

Add a `Body` to `Request`.

## Changes
Added a `Body` to `Request`.
Handle `Body` in the request.
Added `bodyType` and `bodyContent` flags.
Added body information in parsed fields.
Updated all related tests.
Added some tests.

## Notes
I think it is easier to look directly at "files changed" than commit by commit to review.

### DIFF
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -28,6 +28,9 @@ var (
 	timeout       time.Duration // Timeout for each HTTP request
 	interval      time.Duration // Minimum duration between two groups of requests
 	globalTimeout time.Duration // Duration of test
+	method        string        // HTTP request method
+	bodyType      string        // HTTP body type ("application/json" accepted)
+	bodyContent   string        // HTTP request body content as JSON
 )
 
 var defaultConfigFiles = []string{
@@ -54,6 +57,10 @@ func parseArgs() {
 	flag.DurationVar(&interval, "interval", 0, "Minimum duration between two non concurrent requests")
 	// global timeout
 	flag.DurationVar(&globalTimeout, config.FieldGlobalTimeout, 0, "Max duration of test")
+	// body type
+	flag.StringVar(&bodyType, config.FieldBodyType, "", "HTTP body type")
+	// body content
+	flag.StringVar(&bodyContent, config.FieldBodyContent, "", "HTTP request body content as JSON")
 
 	flag.Parse()
 }
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -60,8 +60,7 @@ func parseArgs() {
 	// body type
 	flag.StringVar(&bodyType, config.FieldBodyType, "", "HTTP body type")
 	// body content
-	flag.StringVar(&bodyContent, config.FieldBodyContent, "", "HTTP request body content as JSON")
-
+	flag.StringVar(&bodyContent, config.FieldBodyContent, "", "HTTP request body content as JSON (you must escape double quotes in the object)")
 	flag.Parse()
 }
 
@@ -95,13 +94,16 @@ func parseConfig() (cfg config.Config, err error) {
 		return
 	}
 
-	body := config.NewBody(bodyType, bodyContent)
+	body, err := config.NewBody(bodyType, bodyContent)
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	cliCfg := config.Config{
 		Request: config.Request{
 			Header:  header,
 			Timeout: timeout,
-			Body:    body,
+			Body:    *body,
 		},
 		RunnerOptions: config.RunnerOptions{
 			Requests:      requests,
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -93,11 +93,16 @@ func parseConfig() (cfg config.Config, err error) {
 		return
 	}
 
+	bodyContent, err := config.GetBodyContent(body)
+	if err != nil {
+		fmt.Println(err)
+	}
+
 	cliCfg := config.Config{
 		Request: config.Request{
 			Header:  header,
 			Timeout: timeout,
-			Body:    body,
+			Body:    bodyContent,
 		},
 		RunnerOptions: config.RunnerOptions{
 			Requests:      requests,
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -97,7 +97,7 @@ func parseConfig() (cfg config.Config, err error) {
 		return
 	}
 
-	bodyContent, err := config.ParseBodyContent(body)
+	body, err := config.ParseBody(body)
 	if err != nil {
 		return cfg, nil
 	}
@@ -106,7 +106,7 @@ func parseConfig() (cfg config.Config, err error) {
 		Request: config.Request{
 			Header:  header,
 			Timeout: timeout,
-			Body:    bodyContent,
+			Body:    body,
 		},
 		RunnerOptions: config.RunnerOptions{
 			Requests:      requests,
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -95,7 +95,7 @@ func parseConfig() (cfg config.Config, err error) {
 
 	bodyContent, err := config.GetBodyContent(body)
 	if err != nil {
-		fmt.Println(err)
+		return cfg, nil
 	}
 
 	cliCfg := config.Config{
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -97,7 +97,7 @@ func parseConfig() (cfg config.Config, err error) {
 		return
 	}
 
-	bodyContent, err := config.GetBodyContent(body)
+	bodyContent, err := config.ParseBodyContent(body)
 	if err != nil {
 		return cfg, nil
 	}
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -57,6 +57,8 @@ func parseArgs() {
 	flag.DurationVar(&interval, "interval", 0, "Minimum duration between two non concurrent requests")
 	// global timeout
 	flag.DurationVar(&globalTimeout, config.FieldGlobalTimeout, 0, "Max duration of test")
+	// request method
+	flag.StringVar(&method, config.FieldMethod, "", "HTTP request method")
 	// body type
 	flag.StringVar(&bodyType, config.FieldBodyType, "", "HTTP body type")
 	// body content
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -29,7 +29,7 @@ var (
 	interval      time.Duration // Minimum duration between two groups of requests
 	globalTimeout time.Duration // Duration of test
 	method        string        // HTTP request method
-	body          string        // HTTP body type (format "type:data", only possible type value for the moment: "raw")
+	body          string        // HTTP body type (format "bodyType:bodyContent", only possible bodyType value for the moment: "raw")
 )
 
 var defaultConfigFiles = []string{
--- a/cmd/runner/main.go
+++ b/cmd/runner/main.go
@@ -29,8 +29,7 @@ var (
 	interval      time.Duration // Minimum duration between two groups of requests
 	globalTimeout time.Duration // Duration of test
 	method        string        // HTTP request method
-	bodyType      string        // HTTP body type ("application/json" accepted)
-	bodyContent   string        // HTTP request body content as JSON
+	body          string        // HTTP body type (format "type:data", only possible type value for the moment: "raw")
 )
 
 var defaultConfigFiles = []string{
@@ -59,10 +58,8 @@ func parseArgs() {
 	flag.DurationVar(&globalTimeout, config.FieldGlobalTimeout, 0, "Max duration of test")
 	// request method
 	flag.StringVar(&method, config.FieldMethod, "", "HTTP request method")
-	// body type
-	flag.StringVar(&bodyType, config.FieldBodyType, "", "HTTP body type")
-	// body content
-	flag.StringVar(&bodyContent, config.FieldBodyContent, "", "HTTP request body content as JSON (you must escape double quotes in the object)")
+	// body
+	flag.StringVar(&body, config.FieldBody, "", "HTTP body type")
 	flag.Parse()
 }
 
@@ -96,16 +93,11 @@ func parseConfig() (cfg config.Config, err error) {
 		return
 	}
 
-	body, err := config.NewBody(bodyType, bodyContent)
-	if err != nil {
-		log.Fatal(err)
-	}
-
 	cliCfg := config.Config{
 		Request: config.Request{
 			Header:  header,
 			Timeout: timeout,
-			Body:    *body,
+			Body:    body,
 		},
 		RunnerOptions: config.RunnerOptions{
 			Requests:      requests,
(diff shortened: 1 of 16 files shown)


cmd/runner/main.go
```diff
@@ -54,7 +57,12 @@ func parseArgs() {
 	flag.DurationVar(&interval, "interval", 0, "Minimum duration between two non concurrent requests")
 	// global timeout
 	flag.DurationVar(&globalTimeout, config.FieldGlobalTimeout, 0, "Max duration of test")
-
+	// request method
+	flag.StringVar(&method, config.FieldMethod, "", "HTTP request method")
+	// body type
+	flag.StringVar(&bodyType, config.FieldBodyType, "", "HTTP body type")
+	// body content
+	flag.StringVar(&bodyContent, config.FieldBodyContent, "", "HTTP request body content as JSON (you must escape double quotes in the object)")
```
GregoryAlbouy, line 65: I think a single flag for the two values would be more practical to use, something like
```sh
benchttp -body 'raw:{"id":12}'
benchttp -body file:./my/image.jpg
```
(file is just for the example)
That way we have to parse the value internally, but it avoids having fields required or not based on other fields.
It's not a problem to me that it differs from the config file structure in this case.

config/config.go
```diff
@@ -1,20 +1,44 @@
...
+		err := json.Unmarshal([]byte(bodyContent), &bodyContentMap)
+		if err != nil {
+			return nil, errors.New("bodyContent is not valid json data")
+		}
+	}
+	body.Content = bodyContentMap
+	return &body, nil
+}
+
+var contentTypeValidValues = []string{"application/json"}
```
GregoryAlbouy, line 33: I thought we would treat the input `body` as raw text when we discussed about it 🤔 
It's simpler for us not to know what it is, and I don't think we should perform json validation: what if the user wants to test their endpoint with an invalid json?
Same for the content type: we probably don't want to be smart and dictate it: the user should be able to build its own request without surprises

config/config.go
```diff
@@ -140,13 +176,40 @@ func (cfg Config) Validate() error { //nolint:gocognit
 		inputErrors = append(inputErrors, fmt.Errorf("-globalTimeout: must be > 0, we got %d", cfg.RunnerOptions.GlobalTimeout))
 	}
 
+	if !reflect.ValueOf(cfg.Request.Body).IsZero() {
+		if len(cfg.Request.Body.Content) == 0 {
+			inputErrors = append(inputErrors, fmt.Errorf("-bodyContent: you must provide a value if you have added a bodyType"))
+		}
+		if reflect.ValueOf(cfg.Request.Body.Type).IsZero() {
```
GregoryAlbouy, line 183: Let's avoid calls to `reflect` when we can
```suggestion
		if cfg.Request.Body.Type == "" {
```

test/testdata/config/benchttp.json
```diff
@@ -9,7 +9,14 @@
...
     },
-    "timeout": "2s"
+    "timeout": "2s",
+    "body": {
+    	"type": "application/json",
+    	"content": {
+        "key0": "val0", 
+        "key1": "val1"
+      }
+    }
```
GregoryAlbouy, line 19:
```suggestion
    "body": {
      "type": "application/json",
      "content": {
        "key0": "val0",
        "key1": "val1"
      }
    }
```

config/config.go
```diff
@@ -140,13 +176,40 @@ func (cfg Config) Validate() error { //nolint:gocognit
...
+			inputErrors = append(inputErrors, fmt.Errorf("-bodyType: you must provide a value if you have added a bodyContent"))
+		} else if !contains(contentTypeValidValues, cfg.Request.Body.Type) {
```
GregoryAlbouy, line 185: As far as I remember, that `Type` was supposed to be an indication for us to know how we treat the content: is it raw text or is it a path to a file we should retrieve (or else)

GregoryAlbouy (changes requested):
Ok nice work, this one was a huge piece.

Code is OK to me, but I'm concerned about the specs for the body. The worst is I just realized you strictly followed the linked issue, but the issue is wrong and does not reflect what we decided when we discussed it. Sorry about that :(

config/config.go
moreirathomas: To be honest it was super blurry to me too, even when I wrote the issue (note the number of "may" 😅).

But i agree that just copy pasting the body content provided is a bit easier and gives more freedom to the user (not our fault their requests fail, runs on its machine anyway!).

requester/requester.go
```diff
@@ -108,7 +108,19 @@ func (r *Requester) record(req *http.Request, interval time.Duration) func() {
...
 		sent := time.Now()
 
-		resp, err := r.client.Do(req)
+		// It is necessary to clone the request because one request with a non-nil body cannot be used in concurrent threads
+		// .Clone() does not clone the body, so it is added after with .GetBody()
+		reqClone := req.Clone(req.Context())
+		if req.Body != nil {
+			bodyClone, err := req.GetBody()
+			if err != nil {
+				r.appendRecord(Record{Error: err})
```
moreirathomas, line 117: Could this be a standard error for the package ?

moreirathomas (commented):
@GregoryAlbouy raises some really good points and is suggestions are way better than the fuzzy specs that were written.
I updated the issue.

I guess this is what we could have (pseudo-code):

```go
body := flag.Parse("body", ...)

bodyConf := strings.Split(body, ":")
bodyType, bodyContent :=  bodyConf[0], bodyConf[1]
switch bodyType {
case "raw":
	bodyContent = stringToBytesBuffer(bodyContent) // raw json
case "file":
	bodyContent = fileContentToBytesBuffer(bodyContent) // file path
}

// whenever inside Requester
// body content is bytes.Buffer, don't care more than that about it
req := http.NewRequest(cfg.Haders.Method, cfg.URL, cfg.Body.Content)
```

requester/requester_internal_test.go
```diff
@@ -43,7 +48,9 @@ func TestRun(t *testing.T) {
 			label: "return ErrConnection early on connection error",
 			req: New(config.Config{
 				Request: config.Request{
-					Timeout: 0,
+					Method:  "POST",
+					Timeout: 1,
```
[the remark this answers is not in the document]
c1-ra: Ah yes, it's because of the host that the error should indeed be returned.
I don't understand why adding a body changes anything though. I have retried and there is no `ErrConnection` raised with `Timeout: 0` (locally at least). I got one record which is a success.

cmd/runner/main.go
GregoryAlbouy: Yep that the idea!

> `	// out of scope?`

Yes let's not bother with files and consider it an eternal _future improvement_ 😛

c1-ra:
Thank you for your reviews and correcting the issue!
I have reviewed most of the code.
It became much simpler, so I think it is for the better :)

Just one question: I have used a string for `Request.Body`, @moreirathomas, you were recommending to cast it as a `[]byte`, was there a special reason for this?
I think it is quite simple to have a string and use `strings.NewReader()` to pass it to `http.NewRequest`, but I am not really comfortable with the `[]byte` type so maybe I am not seeing one of its advantages.

config/file/parse.go
```diff
@@ -108,14 +109,13 @@ func parseRawConfig(raw unmarshaledConfig) (config.Config, error) { //nolint:goc
 		fields = append(fields, config.FieldGlobalTimeout)
 	}
 
-	if bodyType := raw.Request.Body.Type; bodyType != nil {
-		cfg.Request.Body.Type = *bodyType
-		fields = append(fields, config.FieldBodyType)
+	body, err := config.GetBodyContent(raw.Request.Body)
+	if err != nil {
+		fmt.Println(err)
```
moreirathomas, line 114: Why do we print but do not handle?

config/file/parser.go
```diff
@@ -192,10 +192,7 @@ type unmarshaledConfig struct {
 		QueryParams map[string]string   `yaml:"queryParams" json:"queryParams"`
 		Header      map[string][]string `yaml:"header" json:"header"`
 		Timeout     *string             `yaml:"timeout" json:"timeout"`
-		Body        struct {
-			Type    *string                 `yaml:"type" json:"type"`
-			Content *map[string]interface{} `yaml:"content" json:"content"`
-		} `yaml:"body" json:"body"`
+		Body        string              `yaml:"body" json:"body"`
```
moreirathomas, line 195: I think @GregoryAlbouy suggested something like 

```yml
body:
  - type: "raw"
  - content: '{"key":"value"}'
```

This is way easier however. I think it's good enough. Everyone ok ?

requester/error.go
```diff
@@ -11,4 +11,6 @@ var (
 	ErrConnection = errors.New("connection error")
 	// ErrReporting is returned when the Requester fails to send the report.
 	ErrReporting = errors.New("reporting error")
+	// ErrRequestBody is returned when the Requester fails to get the body of the request to copy it
+	ErrRequestBody = errors.New("request body error")
```
moreirathomas, line 15:
```suggestion
	ErrCloneBody = errors.New("error cloning the request body")
```

moreirathomas (commented):
Some suggestion regarding docs and `GetBodyContent` implem.
Nice otherwise 👍 

I also think we could go for something like in https://github.com/benchttp/runner/pull/49#discussion_r805320873 for the body.
```go
cfg := config.Config{
	Request: config.Request.WithBody(raw),
	// ...
}
```

But this may be out of scope because it deals with error handling and streamlining the api. So let's put that aside for now !

moreirathomas:
> Just one question: I have used a string for `Request.Body`, @moreirathomas, you were recommending to cast it as a `[]byte`, was there a special reason for this? I think it is quite simple to have a string and use `strings.NewReader()` to pass it to `http.NewRequest`, but I am not really comfortable with the `[]byte` type so maybe I am not seeing one of its advantages.

Because `http.NewRequest` requires a Reader and my mind went directly to `bytes.Buffer` for that. But you're right with `strings.NewReader`, it's convenient  as we already have the string !
This may change when reading the file holding the raw body (`"file:path/to/file.txt"`) as `os.ReadFile` returns a slice of bytes. But it works perfect for now !

config/file/parser.go
GregoryAlbouy: Yep, the condensed syntax was intended for CLI usage only because of its specific constraints, in config files it makes more sense to use the explicit one.

requester/requester_internal_test.go
```diff
@@ -29,7 +32,9 @@ func TestRun(t *testing.T) {
 			label: "return ErrRequest early on request error",
 			req: New(config.Config{
 				Request: config.Request{
+					Method:  "POST",
 					Timeout: 0,
+					Body:    validBody,
```
GregoryAlbouy, line 37: Is it necessary to add `Method` and `Body` in all these tests?

requester/error.go
GregoryAlbouy: Wouldn't it be confusing for a user to get an error _cloning the request body_? 🤔

requester/requester_internal_test.go
c1-ra: I guess that no :sweat_smile: 
I have just let it in the happy path to have a full one.

requester/requester.go
```diff
@@ -108,7 +108,19 @@ func (r *Requester) record(req *http.Request, interval time.Duration) func() {
...
+		// .Clone() does not clone the body, so it is added after with .GetBody()
+		reqClone := req.Clone(req.Context())
+		if req.Body != nil {
+			bodyClone, err := req.GetBody()
+			if err != nil {
+				r.appendRecord(Record{Error: ErrRequestBody})
+				return
+			}
+			reqClone.Body = bodyClone
+		}
```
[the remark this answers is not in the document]
c1-ra: Ah yes, it is better to start the timing after the cloning!

requester/error.go
c1-ra: I have let it as it is for the moment, it should be easily modifiable if we want to later!

test/testdata/config/benchttp.json
```diff
@@ -9,7 +9,11 @@
       "key0": ["val0", "val1"],
       "key1": ["val0"]
     },
-    "timeout": "2s"
+    "timeout": "2s",
+    "body": {
+      "type": "raw",
+      "content": "{\"key0\":\"val0\",\"key1\":\"val1\"}"
```
c1-ra, line 15: I have searched quite a moment if I could do something to avoid escaping these double quotes here, but I don't think it is easily possible.
Moreover, we want a string here (and not a map for example) in case the user wants to input totally wrong json too, right?

config/file/parse_test.go
```diff
@@ -76,6 +77,8 @@ func TestParse(t *testing.T) {
 	t.Run("happy path for all extensions", func(t *testing.T) {
 		for _, ext := range supportedExt {
 			expCfg := newExpConfig()
+			fmt.Println("EXPCONFIG")
+			fmt.Println(expCfg.Request.Body)
```
GregoryAlbouy, line 81: oops

GregoryAlbouy (commented):
Thanks for adding the config file syntax 🙏 

Spotted some oopsies, otherwise we're good to go 🚀

test/testdata/config/benchttp.json
moreirathomas: JSON syntax requires it. It's fine i think. And yes we'd like to allow the user to pass invalid json !

requester/requester.go
```diff
@@ -154,3 +149,18 @@ func (r *Requester) appendRecord(rec Record) {
 		r.numErr++
 	}
 }
+
+// helpers
+
+// Request.Clone() method from net/http packacke does not clone the body, so it is added after with .GetBody()
```
moreirathomas, line 155: nitpick, but kinda related to idiomatic/conventional go:

Unexported methods are helpers, there's no real need to point that out. Also if a function is documented, the documentation explains what the function does :

```go
// cloneRequest fully clones a http.Request by also cloning the body via Request.GetBody.
func cloneRequest(req *http.Request) (*http.Request, error)
```

Rule of thumb is function documentation should be written thinking about how it will appear in godocs (command `make docs`).

config/file/parse.go
```diff
@@ -109,11 +111,8 @@ func parseRawConfig(raw unmarshaledConfig) (config.Config, error) { //nolint:goc
 		fields = append(fields, config.FieldGlobalTimeout)
 	}
 
-	body, err := config.GetBodyContent(raw.Request.Body)
-	if err != nil {
-		fmt.Println(err)
-	}
-	if body != "" {
+	body := config.Body{Type: raw.Request.Body.Type, Content: []byte(raw.Request.Body.Content)}
+	if !reflect.DeepEqual(body, config.NewBody("", "")) {
```
moreirathomas, line 115: `reflect` is heavy, here we're checkin if the type is not empty string and the content is not empty slice.

```suggestion
	if body.Type != "" && len(body.Content) > 0 {
```

config/file/parse_test.go
c1-ra: Ohlala, I did the same at Btf today... I should add a "look for all "print()" and remove them before pushing" post-it on my computer -_-

requester/requester.go
c1-ra: Thanks! Duly noted :)

GregoryAlbouy (approved):

PR #44, opened 2022-02-06
feat: implement config option header

GregoryAlbouy:


moreirathomas (approved):
Changes to override only the values when provided by cli seem perfect 👍

I'll approve for now because it is the first ready between this and #43. But some sync will be needed regarding content-type and content-length headers. In fact, it will required as much specs definition before coding than actually syncing the changes.

GregoryAlbouy:
> But some sync will be needed regarding content-type and content-length headers. In fact, it will required as much specs definition before coding as actually syncing the changes.

Regarding content-type/body specs, what we agreed on (orally) was:
- we let the user build its own (potentially wrong) requests because:
  1. we don't want to be opinionated about what a request should be
  2. we want to allow incorrect requests as it's a valid endpoint benchmarking use case
- as a consequence we don't attempt to set the content-type based on the body, only use the headers as set by the user and the specified body as bytes without further treatment. It's also much simpler to implement and more performant.

c1-ra (approved):
Looks good to me!

config/config.go
```diff
@@ -93,6 +113,8 @@ func (cfg Config) WithURL(rawURL string) Config {
 // New returns a Config initialized with given parameters. The returned Config
 // is not guaranteed to be safe: it must be validated using Config.Validate
 // before usage.
+//
+// TODO: remove this function: https://github.com/benchttp/runner/pull/40/
```
moreirathomas, line 117: What's up with that ? Wasn't `config.New` not used anymore ?

GregoryAlbouy: Well... As we can see in #40, I forgot to actually remove it 😬

PR #45, opened 2022-02-12
feat(ux): implement live feedback

GregoryAlbouy:
## Description

Add live feedback from `requester.Requester` while it's running in the CLI.

benchttp-live-feedback

## Changes

- Make `dispatcher.Dispatcher.Run` return context errors instead of discarding them, so we can retrieve that information in `requester.Requester.Run`


dispatcher/dispatcher.go
```diff
@@ -59,7 +67,7 @@ func (d dispatcher) Do(ctx context.Context, maxIter int, callback func()) error
 	}
 
 	wg.Wait()
-	return nil
```
c1-ra, line 62: [question]
Why were we always returning a nil error before? :thinking:

moreirathomas (approved):
Nice and fancy 🌈 I love it

dispatcher/dispatcher.go
c1-ra: Ok, thanks, I was surprised we returned anything in this case, but we're not doing it anymore anyways!

requester/requester.go
```diff
@@ -77,7 +80,13 @@ func (r *Requester) Run() (Report, error) {
 
 	defer cancel()
 
-	if err := dispatcher.New(numWorker).Do(ctx, maxIter, r.record(req, interval)); err != nil {
+	r.start = time.Now()
+	go r.refreshState()
+
+	switch err := dispatcher.New(numWorker).Do(ctx, maxIter, r.record(req, interval)); err {
+	case nil, context.Canceled, context.DeadlineExceeded:
```
c1-ra, line 87: Why is `case nil` returning `r.end(err)` and not the same thing as default?

GregoryAlbouy: Here `default` is the case of an _real_ error that prevents the correct execution of `Do`, so it returns an empty returns and the error.
The three values in this `case` L87 are expected and do not prevent the run (`context.DeadlineExceeded` correspond to `globalTimeout` reached for instance), so ihere we end the process the regular way via `r.end`
Then we only use the value of the error to display the status in the CLI: https://github.com/benchttp/runner/blob/71013d338af1dea57966dbd4029b241e28c66bc6/requester/print.go#L73-L86

Is it clearer? 🤔

c1-ra: Ok, thanks for the explanation!
It is clearer!

c1-ra (approved):
Nice! :+1:

c1-ra (approved):
:+1: :rainbow:

PR #49, opened 2022-02-12
dev: refactor config responsibilities

GregoryAlbouy:
## Description

The goal of this PR is to reevaluate and refactor the config flow, especially regarding its use by `requester`.

Having a unique `config.Config` entity worked well until now, because it represented everything `requester` needs, and nothing more.
With the further implementations of output handling, this is not gonna be the case anymore.
Because `config.Config` is won't represent the config for `requester` anymore, we don't want `requester` to depend on it.

So here we remove all couplings between `requester` and `config`, so that `requester` becomes a fully independent module, as we did earlier with `dispatcher`.

## Changes

- Rename/Reorder:
  - `config.Config` ➡️  `config.Global`
  - `Config.RunnerOptions` ➡️  `config.Runner`
  - `Config.Request.Timeout` ➡️  `config.Runner.RequestTimeout`
  - `-timeout` ➡️  `-requestTimeout`
- Dependencies
  - `config` and `requester` fully uncoupled
  - `requester` accepts a `requester.Config` interface
  - `config.Global` implements `requester.Config`

## Notes

- 90f16702e5f651bdee2c8748fe0ae59e8fc8e86b`Uncouple packages config and requester` provides a solution to uncouple `config` and `requester` using an **interface** `requester.Config`. It's discussable API-wise, there's an alternative using a third package `internal/adapter` to handle the conversion (note: the conversion being trivial, the only purpose of this package is to keep both entities uncoupled). PR is here: #50 
  EDIT: Actually I think we should just drop that commit and that's it. The mentioned PR is basically the same thing as the preceeding commit, except the conversion is done in a separate package instead of main. A bit overkill given how trivial it is.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -25,23 +25,23 @@ type Runner struct {
 	GlobalTimeout  time.Duration
 }
 
-// Config represents the configuration of the runner.
-// It must be validated using Config.Validate before usage.
-type Config struct {
+// Global represents the global configuration of the runner.
+// It must be validated using Global.Validate before usage.
+type Global struct {
 	Request Request
 	Runner  Runner
 }
 
 // String returns an indented JSON representation of Config
 // for debugging purposes.
-func (cfg Config) String() string {
+func (cfg Global) String() string {
 	b, _ := json.MarshalIndent(cfg, "", "  ")
 	return string(b)
 }
 
 // HTTPRequest returns a *http.Request created from Target. Returns any non-nil
 // error that occurred.
-func (cfg Config) HTTPRequest() (*http.Request, error) {
+func (cfg Global) HTTPRequest() (*http.Request, error) {
 	if cfg.Request.URL == nil {
 		return nil, errors.New("empty url")
 	}
@@ -61,7 +61,7 @@ func (cfg Config) HTTPRequest() (*http.Request, error) {
 // Override returns a new Config based on cfg with overridden values from c.
 // Only fields specified in options are replaced. Accepted options are limited
 // to existing Fields, other values are silently ignored.
-func (cfg Config) Override(c Config, fields ...string) Config {
+func (cfg Global) Override(c Global, fields ...string) Global {
 	for _, field := range fields {
 		switch field {
 		case FieldMethod:
@@ -85,7 +85,7 @@ func (cfg Config) Override(c Config, fields ...string) Config {
 	return cfg
 }
 
-func (cfg *Config) overrideHeader(newHeader http.Header) {
+func (cfg *Global) overrideHeader(newHeader http.Header) {
 	if newHeader == nil {
 		return
 	}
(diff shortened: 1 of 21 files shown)


config/config.go
```diff
@@ -100,7 +101,7 @@ func (cfg *Config) overrideHeader(newHeader http.Header) {
 // WithURL sets the current Config to the parsed *url.URL from rawURL
 // and returns it. Any errors is discarded as a Config can be invalid
 // until Config.Validate is called. The url is guaranteed not to be nil.
-func (cfg Config) WithURL(rawURL string) Config {
+func (cfg Global) WithURL(rawURL string) Global {
```
moreirathomas, line 104: As `Config.Request` is now a standalone (kinda), would it be better if `WithURL` is its method ? 

 ```suggestion
func (r Request) WithURL(rawURL string) Request {
```

And in main:

```go
	cliCfg := config.Global{
		Request: config.Request{
			Header: header,
		}.WithURL(uri),
		Runner: config.Runner{
			// ...
		},
	}
```

config/config.go
```diff
@@ -11,37 +11,38 @@ import (
...
-func (cfg Config) String() string {
+func (cfg Global) String() string {
 	b, _ := json.MarshalIndent(cfg, "", "  ")
 	return string(b)
 }
 
 // HTTPRequest returns a *http.Request created from Target. Returns any non-nil
 // error that occurred.
-func (cfg Config) HTTPRequest() (*http.Request, error) {
+func (cfg Global) HTTPRequest() (*http.Request, error) {
```
moreirathomas, line 45: Same thing here, this may be a method of `config.Request`. What do you think ?

requester/requester.go
```diff
@@ -57,21 +64,16 @@ func New(cfg config.Config) *Requester {
 
 // Run starts the benchmark test and pipelines the results inside a Report.
 // Returns the Report when the test ended and all results have been collected.
-func (r *Requester) Run() (Report, error) {
-	req, err := r.config.HTTPRequest()
-	if err != nil {
-		return Report{}, fmt.Errorf("%w: %s", ErrRequest, err)
-	}
-
+func (r *Requester) Run(req *http.Request) (Report, error) {
```
moreirathomas, line 67: Won't having the same request passed by reference for each dispatch iteration be a problem?
We can easily copy the request value by giving `Requester` access to `config.Global.HTTPRequest()`. We can add this method to `request.Config` interface possibly.

moreirathomas (commented):
It seems like an improvement to me ! `requester` declare what it needs and `config` provides through `main`.

One question about the owner of config transformation methods, and one thought about possible reference/value bug prone pattern.

moreirathomas:
Regarding `Uncouple packages config and requester` via interface vs #50, I think that at the moment the adapter is overkill. It only serves to interface `config` to `requester` directly. The literal interface does the same and let `config` handle how it does it itself.
This pattern may be useful later if more interfacing options come up, for now i think its a bit noisy for little added value.

But in the end go for whatever you think is best for now :)

requester/requester.go
moreirathomas: `Clone` also works ! Good to me 👌

GregoryAlbouy:
> Regarding `Uncouple packages config and requester` via interface vs #50, I think that at the moment the adapter is overkill.

Agreed. Don't you think the interface is overkill too? We could just have a `requester.Config` struct main make the trivial conversion in `main`, which is equivalent to reverting the last commit of the current PR.

moreirathomas:
> https://github.com/benchttp/runner/blob/b857edbe3daa05b75d8ff0da5c902bb7b8bab8eb/cmd/runner/main.go#L80

Ah right ! Not super sexy but does the job very well, until we have a big difference of structure between `config.Runner` and `request.Config` we can keep the direct conversion. And maybe go back to an interface if the conversion becomes more complex ?

c1-ra (approved):
Nice work, it's a lot!
Seems ok for me :)

PR #57, opened 2022-02-17
feat(output): user-defined template

GregoryAlbouy:
### Example

<details>
<summary>Config</summary>
<pre lang="yml">
request:
  method: GET
  url: http://localhost:9999
  queryParams:
    delay: 200ms
runner:
  requests: 10
  concurrency: 2
  requestTimeout: 1s
  globalTimeout: 10s
output:
  out: [stdout]
  template: |
    {{ .Report.Length }}/{{ .Metadata.Config.Runner.Requests }} requests
    {{ .Report.Fail }} errors
    ✔︎ Done in {{ .Report.Duration.Milliseconds }}ms.
</pre>
</details>

<details>
<summary>Output</summary>
<pre>
→ Summary
10/10 requests
0 errors
✔︎ Done in 1015ms.
</pre>
</details>

### DIFF
--- a/output/output.go
+++ b/output/output.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/benchttp/runner/ansi"
@@ -102,9 +103,13 @@ var _ export.Interface = (*Output)(nil)
 
 // String returns a default summary of an Output as a string.
 func (o Output) String() string {
+	if o.Metadata.Config.Output.Template != "" {
+		return o.applyTemplate()
+	}
+
 	line := func(name string, value interface{}) string {
-		const pattern = "%-18s %v\n"
-		return fmt.Sprintf(pattern, name, value)
+		const template = "%-18s %v\n"
+		return fmt.Sprintf(template, name, value)
 	}
 
 	msString := func(d time.Duration) string {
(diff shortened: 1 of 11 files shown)


output/output.go
```diff
@@ -112,6 +117,18 @@ func (o Output) String() string {
...
+	if err != nil {
+		return err.Error()
+	}
+	var b strings.Builder
+	if err := t.Execute(&b, o); err != nil {
+		return err.Error()
+	}
+	return b.String()
+}
+
```
moreirathomas, line 131: As `Output.String()` is called only when we export the final result, an error inside the template given by the user would only happen here.

Thus everything when fine but there's nothing to show because the template is broken.

Could we use the default if an error occurs here ?
```go
func (o Output) applyTemplate() (string, error)
```

output/output.go
```diff
@@ -86,9 +87,13 @@ var _ export.Interface = (*Output)(nil)
 
 // String returns a default summary of an Output as a string.
 func (o Output) String() string {
+	if o.Metadata.Config.Output.Template != "" {
+		return o.applyTemplate()
```
moreirathomas, line 91: Related to previous comment:
```go
if o.Metadata.Config.Output.Template != "" {
	if s, err :=  o.applyTemplate(); err == nil {
		return s
	}
	// User provided template is invalid, continue.
}
```

output/output.go
GregoryAlbouy: Ideally we'd manage this via the config validation as suggested in the issue:
> Optional: perform syntax validation before the run in `config.Global.Validate`

But it requires some re-architecturing because it implies to import `output` in `config` 😕 

Until then, the fallback you suggest is a great idea, let's do this!

GregoryAlbouy:
Ran `git rebase --onto main feat/output-strategies` without conflicts.

GregoryAlbouy:
Rebased with conflicts after #61 merge.

moreirathomas (approved):

PR #63, opened 2022-02-18
feat: update default config values

GregoryAlbouy:

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"net/http"
 	"net/url"
 	"time"
 )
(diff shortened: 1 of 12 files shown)


config/default.go
```diff
@@ -9,13 +10,20 @@ var defaultConfig = Global{
 	Request: Request{
 		Method: "GET",
 		URL:    &url.URL{},
+		Header: http.Header{},
 		Body:   Body{},
 	},
 	Runner: Runner{
 		Concurrency:    1,
```
moreirathomas, line 17: I would have added some concurrency by default, that's a selling point

config/default.go
```diff
@@ -9,13 +10,20 @@ var defaultConfig = Global{
...
-		RequestTimeout: 10 * time.Second,
+		RequestTimeout: 5 * time.Second,
 		GlobalTimeout:  30 * time.Second,
 	},
+	// TODO:
+	// Output: Output{
+	// 	Out:      []OutputStrategy{OutputStdout},
+	// 	Template: "",
+	// 	Silent:   false,
+	// },
```
moreirathomas, line 28: You can target `feat/out-strategies` right now :)

moreirathomas (approved):

moreirathomas (approved):

PR #66, opened 2022-02-19
feat: extend config files

GregoryAlbouy:
## Description

Implement option `extends` for config file.
The file parser resolves the path value recursively until root config is found (no `extends`) and merges all configs into a single one, overriding each parent with the child.

~It currently stop at `10` recursions as a simple guard against circular references.~

~A more accurate strategy could be to store each resolved path in a map and stop if one is encountered twice, but I got _the flemme_.~ Done.

### Example

```yml
# .benchttp.yml
request:
  url: http://localhost:9999

# .benchttp.ci.yml
extends: ./.benchttp.yml
output:
  template: '{{ .Report.Duration.Milliseconds }}'
  silent: true

# resulting config
request:
  url: http://localhost:9999
output:
  template: '{{ .Report.Duration.Milliseconds }}'
  silent: true
```

## Notes

- [x] Implement recursive config file parsing
- [x] Write unit tests
- [x] ~Proposal: rename CLI flag `configFile` ➡️ `extends`~


c1-ra:
Nice!
Since it is already a nice-to-have feature, I think to store each path in a map to check if we don't have circular references instead of stopping at the tenth recursion could be a future enhancement, but not necessary right now.

c1-ra (approved):

config/file/parse.go
```diff
@@ -42,41 +44,116 @@ type unmarshaledConfig struct {
...
+	return cfg, nil
 }
 
-// parseRawConfig parses an input raw config as a config.Config and returns it
-// or the first non-nil error occurring in the process.
-func parseRawConfig(raw unmarshaledConfig) (config.Global, error) { //nolint:gocognit // acceptable complexity for a parsing func
-	const numField = 10 // should match the number of config Fields (not critical)
+// parseRawConfig parses an input raw config as a config.Global and returns
+// a parsedConfig or the first non-nil error occurring in the process.
+func parseRawConfig(uconf unmarshaledConfig) (parsedConfig, error) { //nolint:gocognit // acceptable complexity for a parsing func
```
moreirathomas, line 155:
```suggestion
func newParsedConfig(uconf unmarshaledConfig) (parsedConfig, error) { //nolint:gocognit // acceptable complexity for a parsing func
```
?

config/file/parse.go
```diff
@@ -42,41 +44,116 @@ type unmarshaledConfig struct {
...
-	if err != nil {
-		return cfg, errWithDetails(ErrParse, cfgpath, err)
+	return uconf, nil
+}
+
+// parsedConfig holds a parsed config.Global and the list of its set fields.
+type parsedConfig struct {
+	value  config.Global
+	fields []string
+}
```
moreirathomas, line 124:
```go
func (p parsedConfig) rememberField(field string)
```
Custom method instead of local `appendField` inside `parseRawConfig`?

config/file/parse.go
```diff
@@ -42,41 +44,116 @@ type unmarshaledConfig struct {
...
+	if _, exists := seen[*uconf.Extends]; exists {
+		return uconfs, ErrCircularExtends
+	}
+
+	// record file, append config
+	seen[*uconf.Extends] = true
+	uconfs = append(uconfs, uconf)
+
+	// resolve extended config path
+	parentPath := path.Join(path.Dir(cfgpath), *uconf.Extends)
```
moreirathomas, line 89: What does it do if the the paths are mixed between relative and absolute ?

Not that we should handle that case, but at least return an error to the user.

GregoryAlbouy: Nested paths are correctly resolved, absolute path are not handled (it would require us to be opinionated about what is the root).

In such cases, a relevant `ErrNotExists` is returned by `parseFile`

config/file/parse.go
```diff
@@ -48,11 +48,11 @@ type unmarshaledConfig struct {
 // and returns it or the first non-nil error occurring in the process,
 // which can be any of the values declared in the package.
 func Parse(cfgpath string) (cfg config.Global, err error) {
-	rawCfgs, err := parseFileRecursive(cfgpath, []unmarshaledConfig{})
+	uconfs, err := parseFileRecursive(cfgpath, []unmarshaledConfig{})
```
c1-ra: What does `uconf` stand for?

config/file/parse.go
```diff
@@ -48,43 +48,48 @@ type unmarshaledConfig struct {
...
-	// avoid circular references
-	if len(uconfs) == recursionLimit {
-		return uconfs, ErrExtendLimit
+	// avoid infinite recursion caused by circular reference
+	if _, exists := seen[*uconf.Extends]; exists {
+		return uconfs, ErrCircularExtends
 	}
 
+	// record file, append config
+	seen[*uconf.Extends] = true
```
c1-ra: [question] What does `true` stand for here? Are there cases it can be `false`?
I am wondering why we couldn't we just have a slice here instead of a map.

config/file/parse.go
```diff
@@ -48,30 +48,40 @@ type unmarshaledConfig struct {
 // and returns it or the first non-nil error occurring in the process,
 // which can be any of the values declared in the package.
 func Parse(filename string) (cfg config.Global, err error) {
-	uconfs, err := parseFileRecursive(filename, []unmarshaledConfig{}, map[string]bool{})
+	uconfs, err := parseFileRecursive(filename, []unmarshaledConfig{}, set{})
```
c1-ra, line 51: :+1: Even better than a slice!

config/file/parse.go
GregoryAlbouy: Unmarshaled CONFigs

c1-ra (approved):
I had some questions, but nothing blocking a merge, I am ok merging it as it is now.

moreirathomas (approved):

PR #74, opened 2022-02-27
ops: create cross-platforms build script

GregoryAlbouy:
## Description

make-build

## Notes

- Binary size is about `9MB`
- It would be nice if we manage to test at least one binary per OS:
  - checkout this branch
  - run `make build`
  - run `./bin/benchttp_<os>_<arch>` corresponding to your OS and your arch.

### DIFF
--- a/script/build
+++ b/script/build
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-platforms="darwin/amd64 darwin/arm64 linux/386 linux/amd64 linux/arm linux/arm64 windows/386 windows/amd64"
+platforms="darwin/amd64 darwin/arm64 linux/386 linux/amd64 windows/386 windows/amd64"
 
 # clear bin directory
 rm -rf ./bin/*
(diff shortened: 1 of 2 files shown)


script/build
```diff
@@ -0,0 +1,28 @@
...
+for platform in ${platforms}; do
+    ((i++))
+
+    split=(${platform//// }) # split platform by sep "/"
+    goos="${split[0]}"
+    goarch="${split[1]}"
+    output="benchttp_${goos}_${goarch}" # e.g. benchttp_darwin_amd64
+
+    # add .exe to windows binaries
+    [[ "$goos" == "windows" ]] && output="$output.exe"
```
c1-ra, line 18: (I didn't know this syntax at all in Bash!)

c1-ra (approved):
Ok for me! :)

script/build
GregoryAlbouy: Could be a bad practice, I'm rather terrible at bash 😬

moreirathomas: Works because `&&` requires success of previous command. We could also do:

```suggestion
    if [ "$goos" == "windows" ]; then output="$output.exe"; fi
```

(I think so at least 🤔)

GregoryAlbouy: Is it any better than the short-circuit?

moreirathomas: No idea 🤷‍♂️ I have no preference.

GregoryAlbouy: Let's save an additional commit for the planet! 🚀 🌍

moreirathomas (approved):

PR #79, opened 2022-03-01
fix(config): invalidate config if concurrency > requests

GregoryAlbouy:
## Description

We weren't checking `cfg.Runner.Concurrency <= cfg.Runner.Requests` in `cfg.Validate`.
As a consequence we were displaying an obscur error message for the user (from `dispatcher`):

```sh
go run ./cmd/benchttp run -requests 1 -concurrency 2 -url http://a.b.com
# output:
# invalid value: maxIter: must be >= numWorker (2), got 1
```

This PR fixes it. New error message:

```sh
go run ./cmd/benchttp run -requests 1 -concurrency 2 -url http://a.b.com
# output:
# Invalid value(s) provided:
# -concurrency: must be > 0 and <= requests (1), we got 2
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -151,7 +151,6 @@ func (cfg *Global) overrideHeader(newHeader http.Header) {
 // Validate returns the config and a not nil ErrInvalid if any of the fields provided by the user is not valid
 func (cfg Global) Validate() error { //nolint:gocognit
 	inputErrors := []error{}
-
 	appendError := func(err error) {
 		inputErrors = append(inputErrors, err)
 	}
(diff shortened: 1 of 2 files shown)


c1-ra (approved):
Indeed, I missed this check! :+1:

config/config.go
```diff
@@ -166,8 +165,11 @@ func (cfg Global) Validate() error { //nolint:gocognit
 		appendError(fmt.Errorf("-requests: must be >= 0, we got %d", cfg.Runner.Requests))
 	}
 
-	if cfg.Runner.Concurrency < 1 && cfg.Runner.Concurrency != -1 {
-		appendError(fmt.Errorf("-concurrency: must be > 0, we got %d", cfg.Runner.Concurrency))
+	if cfg.Runner.Concurrency < 1 || cfg.Runner.Concurrency > cfg.Runner.Requests {
```
moreirathomas, line 168: Technically having more available threads than requests to run is not a problem.

Example:
- at most 10 concurrent threads
- do 5 requests
- at 5, the for loop exits, the dispatchers doesn't care, not its job to know it will not be used fully

moreirathomas (commented):
As describe in a comment, I don't think that we should enforce a correlation between concurrency and requests.

Also begs the question: would give `-1` for number of threads results in using `GOMAXPROCS` as upper bound ?

config/config.go
```diff
@@ -166,8 +165,11 @@ func (cfg Global) Validate() error { //nolint:gocognit
...
+		appendError(fmt.Errorf(
+			"-concurrency: must be > 0 and <= requests (%d), we got %d",
```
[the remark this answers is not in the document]
GregoryAlbouy: Yep out of scope, we can consider an issue to harmonize error messages though.

config/config.go
GregoryAlbouy: This change is an alignment with how `dispatcher` _already_ works, so we display an insightful error message.
If we want to change that let's do it in a dedicated discussion + issue + PR!

PR #80, opened 2022-03-02
fix: config file is ignored when no cli flag is provided

GregoryAlbouy:

### DIFF
--- a/cmd/benchttp/run.go
+++ b/cmd/benchttp/run.go
@@ -20,10 +20,6 @@ import (
 type cmdRun struct {
 	flagset *flag.FlagSet
 
-	// defaultConfigFiles is a slice of default config files to look up for
-	// by order of priority if none is provided via the -configFile flag.
-	defaultConfigFiles []string
-
 	// configFile is the parsed value for flag -configFile
 	configFile string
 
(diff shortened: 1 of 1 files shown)


cmd/benchttp/run.go
```diff
@@ -82,21 +78,21 @@ func (cmd *cmdRun) execute(args []string) error {
...
 	// first arg is subcommand "run"
 	// skip parsing if no flags are provided
 	if len(args) <= 1 {
 		return []string{}
 	}
 
+	// config file path
+	cmd.flagset.StringVar(&cmd.configFile,
+		"configFile",
+		cmd.configFile,
```
moreirathomas, line 90: I'm curious: is there a way to defer the evaluation of the default file until we are sure the flag was not provided ?
 Because we can totally search for the default file via `configfile.Find` and then override it with cli flag. Making a file system search for nothing.

I know this was already the behavior as `configfile.Find(cmd.defaultConfigFiles)` as the second argument to `flagset.StringVar` would evaluate immediatly, but seeing the refactor makes me think about this.

GregoryAlbouy: Interesting! We can achieve this easily by reversing the logic: first set `cmd.configFile` to the value of `-configFile` (with a default value of `""`) then if `cmd.configFile == ""` we fallback to the default files.

I'll give a try to that approach but I might prefer the current one if the new one turns out to be more complex or less straightforward.

c1-ra: Should we merge as it is and change it in another PR if we have the time to?
I agree that ideally, not searching for the file if we override it would be better, but it does not seem critical to me.

c1-ra (approved):

PR #86, opened 2022-03-07
feat(template): expose basic stats

GregoryAlbouy:
## Description

Expose a template func `stats` that provides access to basic stats computed on the go.
Stats are not computed if not explicitly called, and are always computed at most once.

### Examples

- Simple call
  ```yml
  output:
    template: '{{ stats.Min }}'
  ```
- Multiple calls (computation done only once)
  ```yml
  output:
    template: |
      Min: {{ stats.Min }}
      Max: {{ stats.Max }}
      Mean: {{ stats.Mean }}
  ```


c1-ra (approved):
:+1:

PR #88, opened 2022-03-07
feat(template): implement fail function

GregoryAlbouy:
⚠️ Merge **after** #87 

## Description

Expose template function `fail` that fails the test with `os.Exit(1)` and a given message when called in a template.

### Usage example

Fail if max duration > 200ms

#### Template

```yml
output:
  template: |
    {{- if ge stats.Max.Milliseconds 200 -}}
      {{ fail "TOO SLOW" }}
    {{- else -}}
      OK
    {{- end -}}
```

#### Passing test

```sh
go run ./cmd/benchttp run -url "http://localhost:9999?delay=100ms"

# Output:
# OK
```

#### Failing test

```sh
go run ./cmd/benchttp run -url "http://localhost:9999?delay=300ms"

# Output:
# test failed: TOO SLOW
# exit status 1
```

### DIFF
--- a/output/error.go
+++ b/output/error.go
@@ -9,6 +9,10 @@ var (
 	// ErrInvalidStrategy reports an unknown strategy set.
 	ErrInvalidStrategy = errors.New("invalid strategy")
 
+	// ErrTplFailTriggered a fail triggered by a user using the function
+	// {{ fail }} in an output template.
+	ErrTplFailTriggered = errors.New("test failed")
+
 	errTemplateEmpty  = errors.New("empty template")
 	errTemplateSyntax = errors.New("template syntax error")
 )
--- a/output/report.go
+++ b/output/report.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"text/template"
 	"time"
 
 	"github.com/benchttp/runner/ansi"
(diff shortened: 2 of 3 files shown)


c1-ra (approved):

moreirathomas:
Question: this works like a circuit breaker after the run but before the report is sent to our server/the output json/the stdout that's right? If I reuse your example: 

```yml
output:
  out: [benchttp, json]
  template: |
    {{- if ge stats.Max.Milliseconds 200 -}}
      {{ fail "TOO SLOW" }}
    {{- else -}}
      OK
    {{- end -}}
```

If there's one request that took 200ms or more, exit without posting the report to `/v1/report` nor writing to a json file.

If i understood correctly that's super cool !

---

Only small nitpick : `ErrTplFailTriggered` with `tpl` abbreviation while `errTemplateEmpty` and `errTemplateSyntax` are not 😅

GregoryAlbouy:
The template only concerns `stdout`, so it doesn't short-circuit json/benchttp outputs.
We could do it quite trivially by checking `rep.errTplFailTriggered` before these, but we made the different outputs independent by design I believe we probably should continue to do so. (Because `template` is only for `stdout`, why would it prevent sending report to Benchttp). It's really two different use cases, one is for CI usage and the other for recording/analysing endpoints during dev.

I'll ship the naming change in the next PR!

moreirathomas:
Ok thanks for the precision ! Really good too :)

PR #90, opened 2022-03-08
ci: unit test output.Report internals

GregoryAlbouy:


c1-ra (approved):
Nice!

Just a small [nitpick / suggestion]:
Couldn't we also rename `output.Report.errTplFailTrigger` to `errTemplateFailTriggered` to be more consistent?

Anyway, +7% coverage up is really great :)

output/report.go
```diff
@@ -19,6 +19,13 @@ import (
 	"github.com/benchttp/runner/requester"
 )
 
+// make export functions mockable
+var (
+	exportStdout   = export.Stdout
+	exportJSONFile = export.JSONFile
+	exportHTTP     = export.HTTP
+)
+
```
moreirathomas, line 28: I don't know if it's ok to have code inside the actual package (`output` and not `output_test`) designed to change what functions declared as vars do.

Isn't there a cleaner wait to do this ? Like passing export functions inside `Report` ?

```go
type Exporter interface {
	Stdout()
	JSONFile() error
	HTTP() error
}

type Report struct {
	Benchmark requester.Benchmark
	Metadata  struct {
		Config     config.Global
		FinishedAt time.Time
	}

	log func(v ...interface{})
	exportFuncs Exporter // or whaterver
}
```

moreirathomas (commented):
Nice coverage increase, not sure about export funcs mocking process though.

output/report.go
GregoryAlbouy: After reflexion, we won't go this way because it results in much more code and complexity only designed for tests than the mockable funcs:
- an interface declaration
- a new type declaration that implements the interface
- a complex flow from `Report.Export`: `rep.Export` > `rep.exportHTTP` > `rep.exportFuncs.http`

moreirathomas: I didn't really like the "return a function to restore their initial value". I read it as "there's a change you shot yourself in the foot, especially because test functions are inside the same package not a dedicated `_test` one.

Will be good enough for the time we have 👌

GregoryAlbouy: Well tests are not shipped in the compilation anyway no matter the package, there's no risk it impacts the source code.

And regarding the tests it's a rather common thing, hence `t.Cleanup` / `jest.beforeAll`, etc...

PR #91, opened 2022-03-09
ci: improve coverage of package config

GregoryAlbouy:
## Changes

- Rename some existing tests
- Refactor some existing tests
- Add some doc comments
- Change Global.Validate error messages
- ...

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -10,24 +10,21 @@ import (
 	"time"
 )
 
+// Body represents a request body associated with a type.
+// The type affects the way the content is processed.
+// If Type == "file", Content is read as a filepath to be resolved.
+// If Type == "raw", Content is attached as-is.
+//
+// Note: only "raw" is supported at the moment.
 type Body struct {
 	Type    string
 	Content []byte
 }
 
-// // To return a Body pbject with Body.Content as a string
-// func (b Body) String() string {
-// 	bodyObject := "\"Body\": "
-// 	bodyType := "\"Type\" :\"" + b.Type + "\""
-// 	bodyContent := "\"Content\": \"" + string(b.Content) + "\""
-// 	return fmt.Sprintf("{%s\r\t%s\r\t%s\r}", bodyObject, bodyType, bodyContent)
-// }
-
-func NewBody(bodyType, bodyContent string) Body {
-	var body Body
-	body.Type = bodyType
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -150,54 +147,54 @@ func (cfg *Global) overrideHeader(newHeader http.Header) {
 
 // Validate returns the config and a not nil ErrInvalid if any of the fields provided by the user is not valid
```
c1-ra, line 148: `ErrInvalid` became an `InvalidConfigError`!

config/config_test.go
```diff
@@ -13,85 +15,69 @@ import (
...
-		err := cfg.Validate()
-		if err != nil {
-			t.Errorf("valid configuration not considered as such")
+		if err := cfg.Validate(); err != nil {
+			t.Errorf("unexpected error: %v", err)
 		}
 	})
 
-	t.Run("test invalid configuration returns ErrInvalid error with correct messages", func(t *testing.T) {
+	t.Run("return cumulated errors if config in invalid", func(t *testing.T) {
```
c1-ra, line 40: A small typo here : ~~in~~ -> is

config/config.go
c1-ra, line 148: `ErrInvalid` became `InvalidConfigError`!

c1-ra:
[question]
Looks good, I ust have one question (concerning what had been done before, but also errors):
Since providing an invalid Url to `withURL()` results in setting an empty url, when we use `Validate()` on it, won't the message be "missing url" instead of "invalid url"?

moreirathomas (approved):

GregoryAlbouy:
> [question] Looks good, I just have one question (concerning what had been done before, but also errors, so it could be changer here): Since providing an invalid Url to `withURL()` results in setting an empty url, when we use `Validate()` on it, won't the message be "missing url" instead of "invalid url"?

It results in a weird mix: the result is "invalid url" because we check for `nil`, not _empty_.
But because the resulting url is empty, the error message is not ideal, like:
```txt
url (""): invalid url
     ^^ not the input url
```

I suggest we leave this aside for now 😛

c1-ra (approved):

c1-ra:
I agree, it is far from critical, let's move forward!

PR #92, opened 2022-03-09
feat: user authentication

GregoryAlbouy:
## Description

- Implement new command `benchttp auth`
  - `benchttp auth login` gives an url to retrieve a token and asks the user to deliver it, then saves the value to `~/.config/benchttp/token.txt`
  - `benchttp auth logout` erases the content of `~/.config/benchttp/token.txt`

- Handle auth in `output`
  - return `ErrNoToken` if token not set
  - return an authentication error if token not validated by benchttp server

- Handle auth in command `run`
  - read and set token in mem
  - handle auth error from `output`

### Examples

- Token not set or Benchttp server responding `401`:
  ```sh
  go run ./cmd/benchttp run
  
  # Output:
  # authentification to benchttp server failed, please run "benchttp auth login" and restart the benchmark
  # exit status 1
  ```

- Benchttp server responding `2xx`:
  ```sh
  go run ./cmd/benchttp run
  
  # Output:
  # → Report sent to Benchttp
  ```

## Tasks

- [x] Implement `benchttp auth login`
  - [x] Display link to webapp token page and prompt for its value
  - [x] Store value into file `~/.config/benchttp/token.txt`
- [x] Implement `benchttp auth logout`
  - [x] Empty file `~/.config/benchttp/token.txt` if it exists
- [x] Update `benchttp run` to handle benchttp server auth
  - [x] Retrieve token from `~/.config/benchttp/token.txt`
  - [x] Add it to the report's POST request's header
  - [x] Handle auth error


internal/auth/auth.go
```diff
@@ -0,0 +1,106 @@
...
+	return nil
+}
+
+// SaveToken create a file to the default path and writes the token
+// into it, or returns an error that is either ErrTokenFind or ErrTokenSave.
+func SaveToken(token string) error {
+	// Resolve token path
+	tokenPath, err := TokenPath()
+	if err != nil {
+		return fmt.Errorf("%w: %s", ErrTokenFind, err)
```
c1-ra, line 59: Isn't it a little strange to return `ErrTokenFind`/"token not found" although the token is not already saved anyway?

GregoryAlbouy: Yes 🤷‍♂️

c1-ra (approved):
:+1:

cmd/benchttp/run.go
```diff
@@ -136,17 +138,25 @@ func (*cmdRun) requesterConfig(cfg config.Global) requester.Config {
...
+		return nil
+	}
+	if output.ExportErrorOf(err).HasAuthError() {
+		return errors.New(
+			"authentification to benchttp server failed, " +
+				`please run "benchttp auth login" and restart the benchmark`,
+		)
+	}
+	return err
+}
```
moreirathomas, line 162: Nitpick:
I would have inlined it because when reading `execute` we do not see the explicit full flow and I don't think the extraction provide any benefit on the other hand.

internal/auth/auth.go
moreirathomas: If the file is always `./config/benchttp/token.txt`, keep `ErrTokenFind` with message `"token file does not exist"` ?

internal/auth/auth.go
```diff
@@ -0,0 +1,106 @@
...
+
+// DeleteToken removes the content of the token file.
+func DeleteToken() error {
+	return SaveToken("")
+}
+
+// TokenPath resolves the default path for the token file. It retrieves
+// the user's home directory and joins TokenDir and TokenName to it.
+// If it fails it returns an ErrTokenFind error.
+func TokenPath() (string, error) {
```
moreirathomas, line 91: I think it could be unexported?

internal/auth/auth.go
```diff
@@ -0,0 +1,106 @@
...
+	ErrTokenFind = errors.New("token not found")
+	// ErrTokenRead reports an error reading a token file.
+	ErrTokenRead = errors.New("invalid token")
+	// ErrTokenSave reports an error saving a token file.
+	ErrTokenSave = errors.New("cannot save token")
+)
+
+// ReadToken reads a token from a file and sets UserToken to the retrieved
+// value or returns a non-nil error that is either ErrTokenFind or ErrTokenRead.
+func ReadToken() error {
```
moreirathomas, line 31: If i get it correctly, the flow is :

- setup a string var `UserToken`
- in `execute`, before exporting to server, load the file value inside `UserToken` via `ReadToken`.
- read non-nil `UserToken` inside `output/report` via `r.Header.Set("Authorization", "Bearer "+auth.UserToken)`

Seems convoluted and relying on implicit state.

Suggestion : 

```go
// cmd/benchttp/run.go

func execute() {
	//...
	token, err := auth.ReadToken()
	if err != nil {
		return err
	}
	err = output.New(rep, cfg).Export(token)
	if err != nil {
		return cmd.handleOutputError(err)
	}
	return nil
```

- `main` really passes all the args
- `output` package does not import `internal/auth`

moreirathomas (commented):
Nicely done ! as usual, nitpicks and actually one suggestion for a better passing of args/not relying on implicit assignation of `UserToken`.

internal/auth/auth.go
GregoryAlbouy: Was first implem, not that bad after all.
Note that it implies to store the token in `output` (either package scope or via a `Report.userToken`, because the token add to headers in method `Report.HTTP` that can't accept extra params because it implements `export.HTTPRequester` (adding it results in a silly signature). So in the end the flow is exactly the same: the only difference being _where_ the token is stored: `auth` vs `output`.

cmd/benchttp/run.go
GregoryAlbouy: Benefits are:
- better readability of the main process (only focus on what is does, not how), which is welcomed here because it orchestrates many things
- decreased complexity

cmd/benchttp/run.go
```diff
@@ -44,24 +44,39 @@ func (cmd *cmdRun) init() {
...
 		return err
 	}
 
+	// Retrieve HTTP request for the benchmark generated by the config
 	req, err := cfg.Request.Value()
 	if err != nil {
 		return err
 	}
 
+	// Retrieve user token if necessary (i.e. exports to benchttp server)
```
moreirathomas, line 62: Nice !

internal/auth/auth.go
```diff
@@ -9,60 +9,58 @@ import (
...
 	// ErrTokenRead reports an error reading a token file.
 	ErrTokenRead = errors.New("invalid token")
 	// ErrTokenSave reports an error saving a token file.
 	ErrTokenSave = errors.New("cannot save token")
 )
 
-// ReadToken reads a token from a file and sets UserToken to the retrieved
+// userToken is the retrieved user token from the token file.
+// It must be set using ReadToken.
+var userToken string
```
moreirathomas, line 27: `userToken` is never assigned a value. That or I missed it.

Either way, we should only read the token once, there's no need to cache it (that's what this declaration does + `if userToken != ""` statement.

It think we can delete lines 25-27 and 32-35 (and update the doc of `ReadToken`) and there's no behavior changes.

Correct me if I'm wrong

moreirathomas (commented):
Better flow ! I still have remarks on `userToken`.

internal/auth/auth.go
GregoryAlbouy: Yep correct, the goal was to set `userToken` at the end of `ReadToken` and returns its value, that way the next call would only read its value if already set. But I forgot about it halfway through 😅 

Deleted the whole thing as we don't have a use case for it.

c1-ra (approved):

moreirathomas (approved):

